PR #5, opened 2016-02-15
Fix Scrabble tile at position (0, 0)

sdgluck:
Currently when a Scrabble tile is at position (0, 0) it becomes hidden by the board due to the `zIndex` of a tile being a concatenation of the `x` and `y` coords. This PR enforces a positive `zIndex` value.

Before:

    zIndex: `${x}${y}`

After:

    zIndex: `${x+1}${y+1}`

As well as this, I have added [`cross-env`](https://www.npmjs.com/package/cross-env) to the examples npm `build` script, and updated the `bundle.js`.

Thanks for a great module. :-)


joshwcomeau:
Thanks for the PR! Never noticed that top square was buggy.

Never heard of `cross-env`, happy to include it though :)

PR #6, opened 2016-02-21
Fix browser globals in UMD build

remko:
While quickly trying out this module on a simple webpage, I noticed the browser globals it uses to bring in React and ReactDOM don't correspond to the ones exported by the React builds.


joshwcomeau:
Ack! Thanks for taking care of this. I feel guilty - I never tested the UMD builds D:

remko:
No problem, it's not something one would use in a real project anyway.
Thanks!

Issue #7, opened 2016-02-23
Internal array not getting reset?


bmueller-sykes:
Hi,

Apologies if this is an ill-formed question/issue. I've been toying around with this library, which seems super cool and very smooth, and during one of my tests I've stumbled across something that is either a bad implementation on my part (for which I'm unclear how to resolve) or a bug in the library. I'll paste in my code below, but the gist is this: I created a simple grid, with a "shuffle" button which, well, shuffles the tiles. As a test, I wanted to change the `opacity` of each tile as it was moving, so I set a "moving" flag on the `onStart()` function, and then unset that flag on `onFinish()` (then, based on that flag, I set the `opacity` to 1 or 0.5. The first shuffle works very smoothly, but on subsequent shuffles, it took longer and longer for the `opacity` to reset. So I threw a `console.log` in the `render()` function, and noticed that with each click of the `shuffle` button, the number of renders kept going up, by the same number of tiles that moved each time. So, on a 3x3 grid, it initially rendered 9 times, then 16, then 23, and so on. In the code below, you'll notice that the number of tiles the code thinks it's rendering doesn't change, which is good--it's the total number of renders that keeps going up.

It kind of looks like the library is keeping track of the items that have moved (obviously, not every tile moves on each shuffle, hence the reason it's not 9, 18, 27, etc), and isn't resetting that counter.

So I don't know if this is a feature or a bug. If it's the former, then I'm not sure how to reset it.

Also (bonus feature request), is it possible to have something like an `onAllFinish()` method? I could certainly see cases where I'd like to be able to fire off an event after all the animations are done.

Thanks!!

Code below:

    import React, { Component } from 'react';
    import FlipMove from 'react-flip-move';
    import _ from "lodash";

    export default class GridContainer extends Component {

	constructor (props) {
		super(props);
		
		
		this.config = {
			rows: 3,
			columns: 3,
			tileWidth: 100,
			tileHeight: 100,
			tileMarginRight: 2
		}
		
		this.config.numTiles = this.config.rows * this.config.columns;
		
		this.state = {
			tileData: _.times(this.config.numTiles, (index) => (
				{
					id: index,
					moving: false
				})
			)			
		}
	}

	handleShuffle = (evt) => {
		evt.preventDefault();
		
		this.setState({
			tileData: _.shuffle(this.state.tileData)
		})
	}

	handleMoveStart = (child, node) => {
		
		const tiles = this.state.tileData.slice(); //create new array to mutate
[...]

PR #10, opened 2016-02-24
Remove transitionEnd event listener after animation end

IamNotUrKitty:
TransitionEnd event listeners not removed automatically, after domNodes reordering, u can check it with this code:
```javascript
import React, {Component} from 'react';
import ReactDom from 'react-dom';
import FlipMove from 'react-flip-move';
import _ from 'lodash'

class Main extends Component{
    constructor(props){
        super(props);
        this.state = {
            arr:_.range(3)
        };
        this.count = 0;
    }
    shuffle(){
        this.setState({arr:this.state.arr.reverse()})
    }
    start(){
        console.log('start');
    }
    finish(){
        this.count++;
        console.log(this.count);
    }
    render(){
        let {arr} = this.state;
        let style = {fontSize:'34px', background:'silver', border:'1px solid black'};
        return (<div className="container">
            <FlipMove
                duration={500}
                onFinish={::this.finish}
                onStart={::this.start}
            >
              {arr.map((i,key)=> {
                  return (<div key={i} style={style} onClick={::this.shuffle}>{i}</div>)
              })}
            </FlipMove>
        </div>);
    }
}
ReactDom.render(<Main/>,document.querySelector('#app'));
```
After some clicking ur console output will be something like 

```javascript
2 start
 1
 2
2 start
 3
 4
 5
 6
2 start
 7
 8
 9
 10
 11
 12
2 start
etc
```
So we can remove it manually with removeEventListener.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -121,11 +121,12 @@ class FlipMove extends Component {
     });
 
     if ( this.props.onStart ) this.props.onStart(child, domNode);
-
-    domNode.addEventListener(transitionEnd, () => {
+    let onFinishHandler = () =>{
       domNode.style.transition = '';
       if ( this.props.onFinish ) this.props.onFinish(child, domNode);
-    });
+      domNode.removeEventListener(transitionEnd, onFinishHandler)
+    };
+    domNode.addEventListener(transitionEnd, onFinishHandler);
   }
 
   childrenWithRefs () {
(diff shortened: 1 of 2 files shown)


IamNotUrKitty:
I think this solve [#7](//https://github.com/joshwcomeau/react-flip-move/issues/7)

joshwcomeau:
Amazing! Yeah for whatever reason it never occurred to me that I'd need to remove `transitionend` event listeners.

You just saved me a couple hours of debugging this weekend. Thank you thank you thank you!

Issue #12, opened 2016-02-26
Nested <FlipMove> lists don't animate properly


cameronhimself:
I have a structure like this:

    <FlipMove>
        <div class="section">
            <FlipMove>
                <div class="item">...</div>
                <div class="item">...</div>
                <div class="item">...</div>
            </FlipMove>
        </div>
        <div class="section">
            <FlipMove>
                <div class="item">...</div>
                <div class="item">...</div>
                <div class="item">...</div>
            <FlipMove>
        </div>
    </FlipMove>

(pretend everything has a unique key property)

The items animate correctly, but when a section is moved all of its child items animate as well, making for a confusing jumble of motion.

(Incidentally, in my use case items can be moved between sections. When that happens the item snaps into the new section with no movement animation (the sections do animate to expand/contract, however). I don't really expect this component to support that kind of functionality, but I figured I'd mention it.)

joshwcomeau:
Hi Cameron!

A project I'm working on will require exactly this functionality, and I was hoping the nested FlipMoves would work. I'll take some time over the next few days to investigate, and hopefully there's an easy solution =)

As for having it animate between sections, that is functionality I spent an evening trying to implement. Ultimately it wound up being more complicated than I had hoped (just being able to shove one `<FlipMove>` at the top and then animating anything with a `key` property), and I didn't want to make the codebase so much more unapproachable for a feature nobody would need.

If I get a couple more requests for it, I'll take another crack at it. It would be awesome to have.

IamNotUrKitty:
@cameronhimself  if I correct understand You,  problem with child animation (when section position is changed) can be solved with `shouldComponentUpdate`. Check this [demo](http://codepen.io/iamnoturkitty/pen/YqKgRw).

cameronhimself:
@IamNotUrKitty I haven't tested your solution, but that certainly looks like it would work in my use case. The only problem I see is that it would involve adding a semantically irrelevant `ItemList` component. Not the worst thing in the world, but I'd prefer to have a clean `Section` -> `Item` hierarchy of components if possible. I'll definitely implement your solution if this issue isn't able to be solved. Thank you for looking into it!

joshwcomeau:
@IamNotUrKitty interesting! 

Yeah, that's a good workaround, but really it shouldn't be necessary. I'm gonna see if I can find a way to account for it in the code base, but for now @cameronhimself, I would go with this solution.

joshwcomeau:
[...]

PR #13, opened 2016-02-28
Calculate positions relative to parent node

IamNotUrKitty:
Fix some problems with nested `FlipMove`, reference to [#12](https://github.com/joshwcomeau/react-flip-move/issues/12)


src/FlipMove.js
```diff
@@ -83,10 +98,14 @@ class FlipMove extends Component {
...
+    const relativeBox = {
+      'top':newBox.top - this.parentBox.top,
+      'left':newBox.left - this.parentBox.left
+    };
 
     return [
-      oldBox.left - newBox.left,
-      oldBox.top  - newBox.top
+      Math.abs(oldBox.left) - Math.abs(relativeBox.left),
+      Math.abs(oldBox.top)  - Math.abs(relativeBox.top)
```
joshwcomeau: Quick question: Why is the `Math.abs` necessary? Seems to me like if the old box was -20px from the parent and the new box is 20px, we want it to move by 40px, not 0px?

Thanks so much for the PR =)

IamNotUrKitty:
Yes you right, it not necessary. Totally forgot for what reason i do it. May be for first i try calculate position something like `parentBox.top - boundingBox.top` and get negative values. Now it not necessary for sure.

joshwcomeau:
Makes sense! 

Everything else looks perfect, and the tests pass. Will merge it in if you can update the PR (If I don't hear from you in a bit, I'll merge it in anyway and fix it up in a rebase. So don't worry if you're busy :))

:rocket:

IamNotUrKitty:
Updated.

PR #27, opened 2016-04-13
Change spelling of "accordion" presets

sethkinast:
I know this is pedantic... but the spelling used is not very typical (Google asks "did you mean..." and freedictionary [redirects you](http://www.thefreedictionary.com/Accordian).

I briefly contemplated leaving aliases in so that people using these wouldn't break; I can do that if it would be helpful.

Or if this is altogether too much of a nit we can just toss it :stuck_out_tongue:


joshwcomeau:
Ack! Of the bugs I and others have found, this has the be the most embarrassing =P

Definitely not too much of a nit. Names matter.

The trouble is, this is sort of a breaking change. I don't much like the idea of having aliases either, but I think it's better than making v3.0 "Fixed typo in preset name".

I think the best way to go about it is just to add them at the bottom of `enter-leave-presets`:

```js
export let enterPresets = { ... }
export let leavePresets = { ... }

enterPresets.accordianVertical = enterPresets.accordionVertical;
leavePresets.accordianVertical = leavePresets.accordionVertical;
```

Thanks for taking the time to prepare a PR :)

sethkinast:
That was my plan! I'll do it in a bit if you haven't.

I believe you can actually still leave them as const because the object
reference won't change.

joshwcomeau:
Oh yeah, right.

I just pushed an unrelated change, but it seems as though there's a conflict in the minified dist file. Could I trouble you to rebuild/rebase it? 

I really need to find an alternative to keeping the UMD builds in the repo. They're too problematic.

sethkinast:
Use a prepublish script maybe?

joshwcomeau:
I do build the dist before publishing, which is essential for generating the ES5 `lib` files. Really I should set up a git hook as well, since the dist files are mainly grabbed from github for fiddling in a jsbin, I believe.

Neither would solve this problem, though. Ideally I'd have them in their own out-of-the-way branch, but then I'd have to remember to keep them in sync.

Oh well. Not a terrible tragedy =)

sethkinast:
I see you do use a prepublish, so yeah they would all be in the npm tarball already even if gitignored. The way I get around this for Dust is to grab latest from CDNJS, which happens to host the library. Maybe adding the lib to one of the open CDNs could be a solution!

PR updated.

joshwcomeau:
Yeah, I've thought about moving it to a CDN actually. That's probably the solution.

Thanks again!

PR #54, opened 2016-06-07
Use package.main instead of empty src/index file

tobiasmuehl:
Saves a bit of space on your harddrive :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-flip-move",
   "version": "2.3.0",
   "description": "Effortless animation between DOM changes (eg. list reordering) using the FLIP technique.",
-  "main": "lib/index.js",
+  "main": "lib/FlipMove.js",
   "files": [
     "dist",
     "lib",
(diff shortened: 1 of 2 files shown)


joshwcomeau:
Yeah, so I get that this file structure is a little silly :) There was a technical reason for doing it this way though... Something about a bug with babel, with having ES6 imports in your main file? It looks like I did this way back in [January](https://github.com/joshwcomeau/react-flip-move/commit/7b49a0206dbba926d0e7a912c47006c7659742e6), but sadly I wasn't very clear in the commit _why_ I was doing this.

It's possible that the issue's been resolved (or that there never was an issue and I'm mistaken!), but there isn't really a ton of benefit to changing it, and I don't think I want to fix what isn't broken :)

Also, it might be wise to move unrelated code (the `@propConverter` wrapper and the polyfills import) to that `index.js`, to keep `FlipMove.js` really all about the flipping and the moving.

Thanks for contributing though! =)

Issue #55, opened 2016-06-13
FlipMove does not take CSS scale() into account.


Guuz:
Hello,

If FlipMove is used inside an element that is scaled the animation start positions are not correctly calculated. An easy example of this is to run this code on this page and use the demo functions:
http://joshwcomeau.github.io/react-flip-move/examples/#/shuffle
```js
$$('#main-content')[0].style.transform = 'scale(0.5)';
```

I've tracked the problem down to `getPositionDelta()`. This should fix it on the vertical axis:
```js
return [oldBox.left - relativeBox.left, oldBox.top / SCALE_FACTOR - relativeBox.top / SCALE_FACTOR];
```

I admit the use case is probably not very big. Our application is scaled to maintain the same aspect ratio and element positions on every screen. But there are other use cases to use `scale()`.

I don't see any other way to get the correct values when using CSS scaling. This is how `getBoundingClientRect()` works. Do you see any other way?
**Would it be an option to introduce a prop for for the scale factor?** It should default to `1` of course.

I also tried extending the component to overwrite the `getPositionDelta()`. But because a `PropConverter` is used extending is "broken". When you extend FlipMove you are actually extending PropConverter so you can't access any of FlipMove's internals. Any thoughts on this?

Thanks in advance!

joshwcomeau:
Hi @Guuz!

Yeah, you're not the first to point out that the PropConverter HOC gets in the way. I never really envisioned anyone extending the FlipMove class, and I see problems with the idea; for one thing, I don't have the discipline to guarantee consistency with internal methods. A patch version could break the extension.

That said, I'd probably be really frustrated in your shoes, since you have a not-unreasonable use-case and FlipMove isn't making it easy for you. Sorry about that!

I wonder if there's a way to do it programmatically; Climb up the node tree checking for transform props, and applying any you find to an overall multiplier. It'd be hard to make it dynamic (I guess it could be done on mount, and on update?), and I don't know how performant it'd be with large DOM trees... Animations would also be a problem (if the scale is animated on load, we'd read the mid-transition value without being aware that it'd about to go away).

If you have any thoughts on how to refine this idea, I'd be super interested in it. Could even be its own NPM package, to help any project deal with `getBoundingClientRect` inside scaled nodes.

For adding the prop, I'd like to wait and see if it becomes a more common issue. I'm trying to avoid implementing fixes for rare issues if it comes at the cost of a more complex API (there's already so many more properties than I had originally wanted). 

[...]

PR #59, opened 2016-06-17
Add new prop `getPosition()`

Guuz:
Hi @joshwcomeau!
This is based on our conversation in https://github.com/joshwcomeau/react-flip-move/issues/55.

This is my proposal on how we could fix the scaling issue with a more generic solution than a `scaleFactor` prop. I couldn't quite decide on what to call the prop. I though about using `getRect` or even `getRekt` but I didn't know if you saw the humor in that. ([get rekt](http://www.urbandictionary.com/define.php?term=get%20rekt))

I tried to make the new code fit in with your existing code as much as possible but feel free to change whatever you want or send me notes on what I should change.

Let me know what you think! I would love to be able to use this prop in our application ASAP :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Flip Move uses the [_FLIP technique_](https://aerotwist.com/blog/flip-your-anima
   * [className](https://github.com/joshwcomeau/react-flip-move#classname)
   * [typeName](https://github.com/joshwcomeau/react-flip-move#typename)
   * [disableAllAnimations](https://github.com/joshwcomeau/react-flip-move#disableallanimations)
+  * [getPosition](https://github.com/joshwcomeau/react-flip-move#getPosition)
   * [HTML Attributes](https://github.com/joshwcomeau/react-flip-move#html-attributes)
 * [Gotchas](https://github.com/joshwcomeau/react-flip-move#gotchas)
 * [Changelog](https://github.com/joshwcomeau/react-flip-move#changelog)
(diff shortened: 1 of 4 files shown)


joshwcomeau:
This is beautiful!

Thanks for matching the code style, and providing documentation and a story. More than I could have asked for. And `getPosition` is a good name, I chuckled at `getRekt` :)

The only thing I ask is you squash your commits into 1. A convention I haven't always followed myself but am trying to maintain is 1-commit-per-feature. I think GitHub lets you submit squashed PRs now, but I have yet to try it myself (I just updated the settings to only allow squash merges).

Excited to get this live as well, will publish a release when the PR is updated!

Guuz:
I'll look into squashing. I don't like forcing a push (rewriting history) but i guess in this case it can't hurt.

If i understand this article you should be able to squash while accepting my PR. But i dont see this feature on my own repo either...
https://help.github.com/articles/about-pull-request-merge-squashing/
Does it show up for you?

joshwcomeau:
Appreciate it!

That was my initial thought too, but I believe it's up to the contributor when the PR is created (that way you can give the new squashed merge a comment. Can you try closing/reopening? I changed the settings so it _should_ be the only option.

If not, I'll just merge it and rebase after the fact, so do reopen it either way :)

(also, I was initially really reluctant to rewrite history as well; I found that to keep myself sane on larger features where I felt like the changes were important, I'd keep an unsquashed copy of the branch, so I still had the commit history if I needed it, and master's history is super linear and nice)

Issue #21, opened 2016-04-08
Prop to Determine Pre-set Translations


marcus433:
Regarding "Existing transition/transform properties will be overridden", for the present what if react-flip-move checked for a prop on the children for example `translate3d={[x, y, z]}`. I believe this would be a valid solution for now as the user most likely will know the translations of their elements before hand. It also saves the need to parse existing CSS properties which may lessen the animation performance. For items that already have a `translate3d` prop I propose 2 custom options for the parent container stating what prop to transform from and whether to take into account transforms at all. Let me know what you think, I'd be happy to contribute to the implementation of this!

Also, great work on the library, I absolutely love it.

joshwcomeau:
Hi Marcus!

In general, I'm a fan of keeping as small an API footprint as possible. I like the spirit of what you're proposing, but I don't like the idea of needing to track half a dozen new properties on the children. It's also a bit mind-bending; took me a couple read-throughs to understand what you meant.

I think the ideal solution is just to compare the currently-applied styles, and "diff" between what they are now and what they need to be.

For example, if an item has a user-specified `transform: rotate(10deg)`, and FlipMove determines that the item needs to shift it up by 20px, we could combine them into `transform: rotate(10deg) translate(-20px, 0)`. 

Another example, if the user specifies a 'transform: translate(10px, 0)` and we need to shift it up by 20px,we'd do the math and wind up with `transform: translate(-10px, 0)`.

This would require a mini-parser implementation that is able to understand `translate`, `translateY`, and `translateX`, and get the difference between the currently-set props and our computed what-needs-to-change props. 

As for performance, I'm not too concerned; we actually already are reading/writing to the node's `styles`, without any noticeable problems. That said, I could be totally wrong; this _is_ a performance-sensitive module, so we'd need to test it.

It doesn't strike me as a super-urgent feature, but it'd certainly be nice to have. If you'd like to take a crack at solving this problem, you're more than welcome to!

marcus433:
Alrighty, sounds good! I'll look into it further. I'm quite interested in achieving this as it will allow support for virtual scrolling implementations!

joshwcomeau:
👍 . Let me know if you have any questions about the source code.

marcus433:
Alright sounds good! So far i've got the parser done. Still need to add matrix support however. It can be seen over at [TransformParser](https://github.com/marcus433/TransformParser/tree/master). I plan to have it done and a pull request ready in the next day or 2 :-).

[...]

PR #81, opened 2016-10-11
Feature/maintain container height container size

### LINKED ISSUE #80: Delay container resize on element removal
Hi @joshwcomeau. Firstly, I'm loving this library! Basically achieved everything I was looking for as a simple drop-in replacement for `ul`. No configuration required.

One problem I'm experiencing is that elements are being removed from the dom flow before they animate away:

flipmove-container-jump

In one view we have a list with very large elements. When removing them the entire document beneath will jump up underneath the element as it animates away. Other similar libraries I've seen do this hold the space open until the element disappears and then snap to the new size. I think this is preferred behaviour.

So - is there already an option for this behaviour? If not, would you like me to provide a PR to address this? If you have any tips for how to make this change that would be good. Otherwise I'm pretty confident that I can navigate your (immaculately commented 👍) codebase.

rhys-vdw:
Implementation inspired by discussions in #80. Handles variable margins, various box-sizing values and interruptions (synchronous "leaving" animations).

Added test and storybook page as requested.

Shortcomings/possible extensions:
- Does nothing for horizontal lists.
- Doesn't allow animation of container height.
- Always creates a `div` element - regardless of parent container. (this is technically not valid HTML if the parent is a `ul`, but I didn't sweat it too much because the tests already put `li` elements in a `div` which is also invalid...)

There's obviously more to do here to make it "right" - but for now this solves my immediate use case and feels like a step in the right direction. I called the prop `maintainContainerHeight` (instead of the suggested `maintainContainerSize`) because it doesn't maintain width.

Closes #80

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -236,17 +236,24 @@ class FlipMove extends Component {
         // We need to find the height of the container *without* the padding
         // element. Since it's possible that the padding element might already
         // be present, we first set its height to 0. This allows the container to
-        // collapse down to the size of just its content.
+        // collapse down to the size of just its content (plus container padding
+        // or borders if any).
         this.heightPlaceholder.style.height = 0;
-        const contentHeight = this.props.getPosition(this.parentElement).height;
+        const collapsedHeight = this.props.getPosition(this.parentElement).height;
 
         // Find the distance by which the container would be collapsed by elements
         // leaving. We compare the temporarily available `collapsedHeight` with
         // the previously cached container height.
-        const collapseHeight = this.parentBox.height - contentHeight;
+        const reductionInHeight = this.parentBox.height - collapsedHeight;
 
-        // Update the padding element's height.
-        this.heightPlaceholder.style.height = Math.max(0, collapseHeight) + 'px';
+        // If the container has become shorter, update the padding element's
+        // height to take up the difference. Otherwise set its height to zero for
+        // no effect.
+        if ( reductionInHeight > 0 ) {
+          this.heightPlaceholder.style.height = `${reductionInHeight}px`;
+        } else {
+          this.heightPlaceholder.style.height = 0;
+        }
       }
     }
 
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -233,9 +233,20 @@ class FlipMove extends Component {
         domNode.style.right = leavingBoundingBox.right - cleanedComputed['margin-right'] + 'px';
       });
 
-      const newHeight = this.props.getPosition(this.parentElement).height;
-      const collapseHeight = this.parentBox.height - newHeight;
-      this.leavingPaddingElement.style.height = collapseHeight + 'px';
+      // We need to find the height of the container *without* the padding
+      // element. Since it's possible that the padding element might already
+      // be present, we first set its height to 0. This allows the container to
+      // collapse down to the size of just its content.
+      this.leavingPaddingElement.style.height = 0;
+      const contentHeight = this.props.getPosition(this.parentElement).height;
+
+      // Find the distance by which the container would be collapsed by elements
+      // leaving. We compare the temporarily available `collapsedHeight` with
+      // the previously cached container height.
+      const collapseHeight = this.parentBox.height - contentHeight;
+
+      // Update the padding element's height.
+      this.leavingPaddingElement.style.height = Math.max(0, collapseHeight) + 'px';
     }
 
     const dynamicChildren = this.state.children.filter(
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -104,7 +104,8 @@ class FlipMove extends Component {
         'top':    childBox['top']  - parentBox['top'],
         'left':   childBox['left'] - parentBox['left'],
         'right':  parentBox['right'] - childBox['right'],
-        'bottom': parentBox['bottom'] - childBox['bottom']
+        'bottom': parentBox['bottom'] - childBox['bottom'],
+        'height': childBox['height']
       };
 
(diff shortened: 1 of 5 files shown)


src/FlipMove.js
```diff
@@ -233,6 +231,23 @@ class FlipMove extends Component {
...
+        this.heightPlaceholder.style.height = 0;
+        const contentHeight = this.props.getPosition(this.parentElement).height;
+
+        // Find the distance by which the container would be collapsed by elements
+        // leaving. We compare the temporarily available `collapsedHeight` with
+        // the previously cached container height.
+        const collapseHeight = this.parentBox.height - contentHeight;
+
+        // Update the padding element's height.
+        this.heightPlaceholder.style.height = Math.max(0, collapseHeight) + 'px';
```
joshwcomeau: So it's more verbose, but I think it might be clearer if it was something like:

```js
// First two lines unchanged
this.heightPlaceholder.style.height = 0;
const contentHeight = this.props.getPosition(this.parentElement).height;

const hasContentShrunk = contentHeight < this.parentBox.height;

if ( hasContentShrunk ) {
  const changeInHeight = this.parentBox.height - contentHeight;
  this.heightPlaceholder.style.height = `${changeInHeight}px`;
} else {
  this.heightPlaceholder.style.height = 0;
}
```

I think it's clearer because it's obvious that we're only setting the placeholder's height to the diff when the content has shrunk.

(Also, thanks for showing me that `Math.max` trick; really nice built-in alternative to lodash's `clamp`)

stories/index.js
```diff
@@ -141,6 +141,9 @@ storiesOf('FlipMove', module)
 
     return <HandleEmpty />
   })
+  .add('maintain container height', () => (
+    <Controls maintainContainerHeight={true} style={{border: 'solid 2px magenta'}} />
```
joshwcomeau: Can you add some padding to this story? I checked it locally and it does work :) just wanna make sure the story includes it.

joshwcomeau:
And yeah, I do see an argument for inferring the type of the placeholder based on the parent (a `<li>` when inside a `<ul>`)... You're right that the tests are super sloppy for this project.

I'm happy to contribute that part though, can take care of it after this is merged.

src/FlipMove.js
rhys-vdw: Cool. Changed it a bit. It's not exactly like your code above, but hopefully it's explicit enough.

btw, a common way of clamping a number is like this:

```js
function clamp(value, lower, upper) {
  return Math.min(Math.max(value, lower), upper)
}
```

rhys-vdw:
Thanks @joshwcomeau, made some changes.

joshwcomeau:
@rhys-vdw thanks so much for your work on this feature!

I've made you a collaborator on the repo :)

rhys-vdw:
> I've made you a collaborator on the repo :)

Thanks! I'd love to get a release out soon so I can put these changes into production. Anything you need help with?

joshwcomeau:
@rhys-vdw Already done :) v2.6.0 has both your work and my small tweak to support `<ul>` containers. I'm gonna add the couple changes you've mentioned and release v2.6.1 today.

joshwcomeau:
As for help, you're welcome to contribute as much or as little as you'd like.

I was listening to a podcast with [John-David Dalton](https://github.com/jdalton) (Lodash creator), and he says that once developers have made a significant contribution, he gives them full collaborator rights. I think it's a neat idea, so I'm trying it :)

I sadly don't have too much time to dedicate to this project, aside from bug fixes and  maintenance. There are a few things I'd like to implement eventually though. I think I'm gonna give the new GitHub Projects thing a try, and create some feature tickets. Totally up to you if you feel like tackling any of them or not!

rhys-vdw:
Thanks @joshwcomeau. I try to apply the same mentality. Well, unless otherwise prompted I will probably just fix bugs as they come up. Presently the appearance is acceptable for our app, but there may come a time where I have time to work on container animation.

Btw, I've been thinking about this problem a bit more. Do you think it would have been a better approach to just re-sort the children so that the "leaving" elements sorted to the end, but are positioned by `position: relative` (like the other elements)?

joshwcomeau:
Huh... that's a really clever idea, actually.

I like that it extends the basic "FLIP" principle. Instead of having this special leave process, we're just moving its container to the end, translating it back to its original position, and applying another transform to make it exit.

The only complication is the conflicting transforms; the `translate` required to return it to its original position would need to be merged with the exit transforms. This isn't actually a very hard problem, though - We discussed merging transforms in https://github.com/joshwcomeau/react-flip-move/issues/21, and Marcus433 realized that we can just use the browser's rendering engine to work out the matrix of several transforms.

I suspect that this would improve the codebase by quite a bit :)

rhys-vdw:
Yeah. The main motivation was to get rid of the "placeholder" element, and to reuse the dom flow for positioning everything. Still trying to think of an elegant way to animate the container...

joshwcomeau:
The problem with container animation is AFAIK there's no performant way to do it. 

One idea I had was that we could have a `backgroundColor` prop, create an element of that color that fills the container, behind all the elements, and `transform: scale` it up and down as needed. There are a few issues, though:

- It doesn't handle animating the content below the `FlipMove` container
- It assumes a single solid color works as the background
- It's another "placeholder" element, feels unnecessarily complex.

rhys-vdw:
@joshwcomeau Perhaps you could wrap an entire list in an extra `div`, and modify its height based on the internal list height. So long as it's set to `overflow: visible` it shouldn't make a visual difference. That way the inner list can snap up and down, and the outer can translate from its previous size to the new size. But the actual size would be determined by allowing the `ul` to naturally find its layout.

PR #82, opened 2016-10-12
Support <li> placeholders inside <ul>s

joshwcomeau:


src/FlipMove.js
```diff
@@ -508,16 +508,22 @@ class FlipMove extends Component {
       return React.cloneElement(child, { ref: child.key });
     });
 
+    // If requested, create an invisible element at the end of the list.
+    // Its height will be modified to prevent the container from collapsing
+    // prematurely.
     if ( this.props.leaveAnimation && this.props.maintainContainerHeight ) {
+      // Create a <li> when nested inside a <ul>
+      const placeholderType = this.props.typeName === 'ul' ? 'li' : 'div';
```
rhys-vdw: Also should consider [`ol`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol).

src/FlipMove.js
```diff
@@ -508,16 +508,22 @@ class FlipMove extends Component {
...
-          key='height-placeholder'
-          ref={element => { this.heightPlaceholder = element }}
-          style={{ visibility: 'hidden', height: 0 }}
-        />
+        React.createElement(
+          placeholderType,
+          {
+            key: 'height-placeholder',
+            ref: element => { this.heightPlaceholder = element },
+            style: { visibility: 'hidden', height: 0 },
```
rhys-vdw: Also perhaps we should provide `border: 0, margin: 0, padding: 0` here in case they have styles targeting `li`.

joshwcomeau: :+1:

PR #88, opened 2016-10-31
Prevent error with HTMLElement when not redering from browser

kdemoya:
A change in the latest version is causing server rendering (or testing, in my case) to fail because of `HTMLElement` being undefined.

Added a simple check for that.


joshwcomeau:
Ah! Sorry about that. Thanks so much for taking the time to find and solve the problem :)

joshwcomeau:
Just published v2.6.4 which fixes this issue

kdemoya:
Thanks @joshwcomeau, blazing fast. 👍

PR #90, opened 2016-11-03
minor change to example custom animation in readme

finetype:
I liked this custom animation a lot, but found that it caused weird 'hangs' at the beginning and end of animations when sorting multiple items per transition. By making these changes, the transitions are greatly smoothed out and much more pleasing to the eye. I also changed the format just a little so that its more common usage pattern is a little more obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,26 +172,29 @@ You can supply your own CSS-based transitions to customize the behaviour. Both `
 ![Custom](https://s3.amazonaws.com/githubdocs/fm-custom-rotate-x.gif)
 
 ```js
-<FlipMove
-  staggerDelayBy={50}
-  enterAnimation={{
-    from: {
-      transform: 'rotateX(135deg)'
-    },
-    to: {
-      transform: ''
-    }
-  }}
-  leaveAnimation={{
-    from: {
-      transform: ''
-    },
-    to: {
-      transform: 'rotateX(-120deg)',
-      opacity: 0.6
-    }
-  }}
-/>
+ <FlipMove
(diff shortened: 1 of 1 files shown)


joshwcomeau:
Huh. Shame on me for not actually _testing_ the animation in my README! I think I just plucked some random-ish values, similar to ones being used in the Storybook stories.

Good call on `this.renderRows()` as well.

Surprised to hear that there are weird hangs though; I'm going to have to look into this and see what's up with that. Ideally there should be no possible settings that make it look weird!

finetype:
Thanks a ton for this library, it works beautifully. It'll soon be in use to fix the horribly buggy legacy knockout.js based animation clusterfuck that is in place on thezebra.com. Should be in production within a week.

finetype:
The hanging effect was caused by the stagger delay combined with the opacity for starting the animation be 0.6; that meant things would pop into view at 0.6 opacity, and then sit there like that until it was their turn to animate to full opacity (or vice versa). I don't think it can be blamed on the library, just the particular settings provided. Not 100% sure, but that was my interpretation of what I saw.

joshwcomeau:
:D glad to hear it's proving useful! 

And thanks for letting me know where it's being used! I realized recently that while it gets downloaded thousands of times a month, I don't know of a single site (other than ones I've worked on) that use it in production. Makes me happy to know that more than just hobby sites are using it =)

joshwcomeau:
Right, makes sense. I haven't tested staggered enter animations much, I should see if there's a way to smooth that out

PR #91, opened 2016-11-04
Fix case where there is no childrenData for a child prop

aligon:
I was hitting a case where I had a child in the props' children, that did not have any data in the childrenData. I was having trouble tracking down how it got into that state. I have multiple nested flipmove instances, so that might have something to do with it. I thought I'd try just preventing it from blowing up and everything seemed to work correctly afterwards.

The error was occurring in getRelativeBoundingBox, when childData was undefined.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -393,7 +393,9 @@ class FlipMove extends Component {
     this.props.children.forEach((child) => {
       // It is possible that a child does not have a `key` property;
       // Ignore these children, they don't need to be moved.
-      if (!child.key) {
+      // It is also possible that there is no childrenData for the key;
+      // Also ignore these.
+      if (!child.key || !this.childrenData[child.key]) {
         return;
       }
 


joshwcomeau:
Hi @aligon! Thanks for taking the time to submit a PR :)

I'm a little wary to accept a PR like this, because it's masking what may be a serious issue. `childrenData` is populated from the refs on render, and it only tries to access that data when the component receives new props. By this logic, the error you've encounted should be impossible! So clearly there's a leak somewhere, and it may cause more issues if I just shove a band-aid on it.

That said, I also recognize that it's better for an animation to fail than for the entire component to blow up. 

I'm gonna spend some time this weekend seeing if I can figure it out, and if not I'll accept this PR.

Nested FlipMoves _shouldn't_ affect it at all, but I'll keep an open mind. Do you have any other insights you can share? Stuff like:

* what props did you pass <FlipMove>? 
* what are you using for keys?
* how often were things updating?
* does the error happen on enter, leave, shuffle, or all of the above?
* is this a new issue? I refactored the codebase about 2 weeks ago, so I'm curious if it was working before then or not.

Thanks again!

aligon:
The only way I've found to reproduce it, is to re-order the children of one of the nested instances. Then try to add another item to the outer. After that any attempt to add a new item fails.

I'm passing these props:
- enterAnimation="fade"
- leaveAnimation="fade"
- duration={150}
- easing="ease-in"
- disableAllAnimations={disableAnimation}

The keys are either an UUID while the item is being saved, or an ID from a Database once it has finished saving.

I don't think they are updating too frequently. Should only be with user interaction.

I'm seeing the error on enter.

I don't think it is a new issue. I was a few versions behind when I noticed the issue. I'm still seeing it on 2.6.4.

Let me know if there's any more info I can provide.

Thanks for looking into it.

joshwcomeau:
Sorry for the delay, and thank you for all the info!

I haven't yet had any time to dig into this, but I'm going to spend some time tonight looking into it.

Thanks for your patience!

joshwcomeau:
`v2.6.5` includes this fix. Thank you for your contribution!

I'll continue to try and find the underlying issue, but this should keep the lights on for now :)

PR #92, opened 2016-11-09
When not animating, allow the children to update.

aligon:
If animation is disabled, allow the children to update from the new props.

I am turning animation off for a few different user actions. While it is turned off, any changes to the children are rendered one prop change behind. A forceUpdate triggers it to get up to date.

For example, with animation disabled:
Start with children A, B;
Add child C;
FlipMove will still show A, B;
Add child D;
FlipMove will now show A, B, C;

I think this also resolves pull request 91.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -114,7 +114,10 @@ class FlipMove extends Component {
     // IMPORTANT: We need to make sure that the children have actually changed.
     // At the end of the transition, we clean up nodes that need to be removed.
     // We DON'T want this cleanup to trigger another update.
-    const shouldTriggerFLIP = this.props.children !== previousProps.children;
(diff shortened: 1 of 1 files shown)


joshwcomeau:
Ah, yeah this makes sense.

I believe the solution may require a bit more fine-tuning; the tests are failing and suggest that doing it this way causes their on-screen position to break.

Possible that there was a flaw in the tests that is made obvious by this fix, but I'll need to dig into it.

joshwcomeau:
I believe I found the issue; enter/leave animations were still being triggered when animations were disabled, which was causing the test to fail.

Not sure why the tests were passing before the `this.props` -> `nextProps` switch. The solution was simple enough though!

Thanks again :)

PR #93, opened 2016-11-10
Remove references to nodes that have left

haste:
These two commits fix the following issues:
1) `this.childrenToAnimate` is only reset if you have a `onFinishAll` hook, which makes the list grow forever if you don't define one.
2) Children that have been animated away are still referenced by `this.childrenData`, preventing the browser from garbage collecting them.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -348,10 +348,10 @@ class FlipMove extends Component {
           const [elements, domNodes] = this.formatChildrenForHooks();
 
           this.props.onFinishAll(elements, domNodes);
-
-          // Reset our variables for the next iteration
-          this.childrenToAnimate = [];
         }
+
+        // Reset our variables for the next iteration
+        this.childrenToAnimate = [];
       });
 
       // If the placeholder was holding the container open while elements were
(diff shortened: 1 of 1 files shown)


joshwcomeau:
Ah jeez. Thanks so much for spotting these! I refactored a little too recklessly, it seems.

Very good catch on maintaining `this.childrenData` beyond its removal from the DOM. 

I assume that resetting `this.childrenToAnimate` will need to be moved outside the `onFinishAll` conditional, instead of just being removed entirely? 

I'm at work now but I'll dedicate some time tonight to this.

haste:
Thanks for dedicating some time to look into it. You have a lovely written library. :)

The resetting of `this.childrenToAnimate` was moved outside the `onFinishAll` condition in the commit.

joshwcomeau:
Thanks @haste! This fix has been published in `v2.6.6`.

(Not sure how I didn't see that the resetting was moved!)

joshwcomeau:
@haste I've added you as a collaborator to the repo :) 

Lodash creator JDD has a policy where when users submit a quality PR, he makes them a collaborator. I've started following this advice, and so far it's worked out well!

I don't have any expectations, it's up to you to contribute as much or as little as you'd like. Feel free to open PRs, review other people's PRs, suggest features, etc. Totally fine if you don't want to do anything as well!

haste:
Hah, that's a pretty a smart policy! We'll see what happens in the future. :)

PR #101, opened 2016-11-17
Feature/graceful interrupts

joshwcomeau:
* Shuffle interrupts are now handled gracefully in Chrome & Safari. Firefox remains unchanged, and I haven't tested IE/edge.
* Leave animation interrupts should be fixed. It's still a little glitchy in very elaborate scenarios with very long transition times, but sequentially removing several items in a row works. Miles better than the 'restarts-on-every-change' behaviour currently implemented.
* Update React Storybook

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "react-dom": "0.13.x || 0.14.x || 15.x.x"
   },
   "devDependencies": {
-    "@kadira/storybook": "1.19.0",
+    "@kadira/storybook": "2.30.1",
     "babel": "6.1.18",
     "babel-cli": "6.2.0",
     "babel-core": "6.2.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-flip-move",
-  "version": "2.6.10",
+  "version": "2.7.0-beta1",
   "description": "Effortless animation between DOM changes (eg. list reordering) using the FLIP technique.",
   "main": "lib/index.js",
   "files": [
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -398,6 +398,12 @@ class FlipMove extends Component {
     );
 
     this.props.children.forEach((child) => {
+      // It is possible that a child does not have a `key` property;
+      // Ignore these children, they don't need to be moved.
+      if (!child.key) {
+        return;
+      }
+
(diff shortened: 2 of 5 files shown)


joshwcomeau:
I just published version `2.7.0-beta1` which includes:

* Support for graceful shuffle interrupts in Chrome and Safari. Firefox behaviour is unchanged, and I haven't tried other browsers yet
* Support for graceful enter/exit animations. Fixes the bug where leave animations would restart every time the component updates.

I'd be grateful to anyone who can take a moment and install this beta version, see how it works, help me identify any bugs with the implementation :)

cc @rhys-vdw, @haste, @yp, @whitegrey, @valentinvichnal, @SDSLeon

joshwcomeau:
Found a bug that's causing `runAnimation` to fire way more often than it should. Isn't noticeable in contrived examples but perf suffers in real-world usage. Will keep poking at it :)

joshwcomeau:
@rhys-vdw so, graceful interrupts are working well, but they expose an issue I'd like to fix.

Without specifying a z-index, items are rendered in the order they appear in the DOM. When you shuffle the items, their DOM order changes, which means items in the back instantly snap in front when the transition starts.

I was initially thinking that the solution could be to have a `preserveInitialZIndex` prop that would keep an array of all the `key`s it's seen, and apply z-index based on the index of the key in the array. That way, even when new items are added, they're automatically appended to the array and they can be shuffled without affecting their "stack" position.

The thing is, this would complicate the source code, and it's likely a pretty niche concern (it only really matters when you have very frequent shuffles that interrupt each other).

There has been a few cases where I have some niche thing that I kinda wanna implement, but feel like it doesn't belong in the "core" code. We could implement some kind of addon system that would allow us to do stuff like this:

```js
import FlipMove, { withPreservedZIndex } from 'react-flip-move';

const FlipMoveWithPreservedZIndex = withPreservedZIndex(FlipMove);

const MyComponent = () => (
  <FlipMoveWithPreservedZIndex>
    {...}
  </FlipMoveWithPreservedZIndex>
);
```

To make this work, we'd need to find a way for the parent component to access the refs collected by FlipMove, which I think feels a little hacky, but is an acceptable price to pay for the amount of extensibility it would provide. FlipMove could accept an `applyWithRefs` function, supplied by the HOC, that would be invoked with the refs whenever the list is changed.

Curious to know what you think of this idea!

PR #103, opened 2016-11-20
Relative TOC links

rhys-vdw:
Noticed that the table of contents will navigate away from [npmjs.com/package/react-flip-move](npmjs.com/package/react-flip-move) because the href URLs are absolute.

Also converted a couple of HTML tags to markdown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,36 +26,36 @@ Flip Move uses the [_FLIP technique_](https://aerotwist.com/blog/flip-your-anima
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


joshwcomeau:
:+1: thanks!

I've actually started migrating the docs to a gitbook. I find that the README's getting a little long, so I'd like to have the main README cover features, installation and important notes (like the gotchas), but move API reference and enter/leave descriptions to their own files.

Also I have a question for you about something, but I'll ask it in a more appropriate place :)

PR #110, opened 2016-12-02
removed lodash.omit dependency which shaves off ~6.5kb of min+gzip size

HenrikJoreteg:
Hi Joshua! We've been loving this module thanks for sharing it! I'm doing a bit of bundle analysis of our app and noticed that this included lodash.omit which, as it turns out, also has several sub dependencies and ends up adding a surprising (at least to me) 6.5kb to the gzipped/minified size by my measurements. 

Given that `omit` is fairly simple in concept, I thought it'd may be worthwhile to inline a little `omit` implementation and send it your way for feedback.

Thanks again for this slick module!

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,8 +75,5 @@
     "sinon-chai": "2.8.0",
     "style-loader": "0.13.1",
     "webpack": "1.12.11"
-  },
-  "dependencies": {
-    "lodash.omit": "4.3.0"
   }
 }
--- a/src/prop-converter.js
+++ b/src/prop-converter.js
@@ -12,7 +12,6 @@
  */
 
 import React, { Component, PropTypes } from 'react';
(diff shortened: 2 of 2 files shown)


joshwcomeau:
Hi @HenrikJoreteg!

Yeah, this is great. I had no idea `omit` was so huge!

Could I trouble you to move the omit function into `src/helpers`?

Thanks for the PR!

joshwcomeau:
I'm about to pen a small release, so I decided to include this :) don't worry about moving the omit function, I'll tackle it in a subsequent commit.

Thanks again for your contribution!

HenrikJoreteg:
@joshwcomeau had a busy weekend, didn't get to it in time, anyway... thanks! :)

PR #111, opened 2016-12-04
Fix/106

### LINKED ISSUE #106: Cannot read property 'top' of undefined / 'getBoundingClientRect' of null
When removing adding and removing items the FlipMove breaks with the error:

Uncaught TypeError: Cannot read property 'top' of undefined:
https://github.com/joshwcomeau/react-flip-move/blob/master/src/dom-manipulation.js#L141

Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null:
And https://github.com/joshwcomeau/react-flip-move/blob/master/src/prop-converter.js#L217

You can see these errors being thrown by clicking rapidly on the add / remove buttons using the following example:
https://gist.github.com/ConneXNL/c320db6f8898982f78b1b5cd4dc4cfea

joshwcomeau:
Fixes #106 
(although it's really more of a band-aid than a proper fix)

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -196,7 +196,9 @@ class FlipMove extends Component {
 
         // We need to take the items out of the "flow" of the document, so that
         // its siblings can move to take its place.
-        removeNodeFromDOMFlow(childData);
+        if (childData.boundingBox) {
+          removeNodeFromDOMFlow(childData);
+        }
       });
 
       if (maintainContainerHeight) {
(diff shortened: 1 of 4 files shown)


src/FlipMove.js
```diff
@@ -196,7 +196,9 @@ class FlipMove extends Component {
 
         // We need to take the items out of the "flow" of the document, so that
         // its siblings can move to take its place.
-        removeNodeFromDOMFlow(childData);
+        if (childData.boundingBox) {
+          removeNodeFromDOMFlow(childData);
+        }
```
joshwcomeau: The obvious solution here is to just recalculate the bounding box if it's missing. Sadly, this doesn't work, for reasons I don't understand or have time to dig into; it does recalculate the bounding boxes, but it causes the elements to reshape in weird ways.

PR #116, opened 2016-12-15
isAnimationDisabled should use nextProps

willemx:
in componentWillReceiveProps, isAnimationDisabled should use nextProps in stead of this.props


joshwcomeau:
Almost missed this one!

Looks good to me. Thanks for taking the time to fix :)

PR #119, opened 2016-12-22
Fix weird animation when inside component get change by set while animating

chrisza4:
Right now, if you try to change some component inside FlipMove while animation is running, it will cause a weird animation as stated in "Known Issues"

This happen because when component inside FilpMove changes, it will try to re-calculate animating components in `runAnimation` method again via `componentDidUpdate` and store into `remainingAnimations`. And if animation is already running, it will doubled up `remainingAnimations` count.

I workaround that by instead of comparing children in each prop directly, I compare children key equality. So if the children is the same set, FlipMove should not rerun animation.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -21,7 +21,7 @@ import {
   updateHeightPlaceholder,
   whichTransitionEvent,
 } from './dom-manipulation';
-import { arraysEqual } from './helpers'
+import { arraysEqual } from './helpers';
 
 const transitionEnd = whichTransitionEvent();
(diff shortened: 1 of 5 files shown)


joshwcomeau:
Hi @chrisza4,

Sorry it took me so long to see this! Not sure how I missed it. 

This seems like a good idea to me; I'm just a little worried about unintended consequences of this change.

Thanks for allowing me to add commits to this PR; I'm gonna play with it, and make a couple tweaks.

Also, thanks for your contribution :)

joshwcomeau:
Published in `v2.7.4`

PR #121, opened 2017-01-09
Make sure the transition end handler gets called, even if the browser event does not fire.

aligon:
I was hitting a case where after passing a new set of children with some removed, the items that were suppose to be removed were still present. Upon further inspection the remainingAnimations property was still 1. Calling triggerFinishHooks manually seems to fix it.

My assumption was that somehow one item wasn't getting the transition end handler called, so I tried making sure that we call transitionEndHandler.

With that change I'm no longer seeing the original issue, granted it was happening sporadically.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -314,15 +314,20 @@ class FlipMove extends Component {
 
   bindTransitionEndHandler(child) {
     const { domNode } = this.childrenData[child.key];
+    const {duration} = this.props;
+    let ended = false;
 
     // The onFinish callback needs to be bound to the transitionEnd event.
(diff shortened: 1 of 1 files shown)


joshwcomeau:
Hey @aligon,

Sorry I didn't get to this sooner!

So, this PR addresses a symptom, instead of the cause. I don't want to get into the habit of merging code like this in, because it's super unmaintainable.

If you can create a reliable reproducible case on jsbin or similar, I can spend some time seeing if I can diagnose/fix the issue.

That said, if this solution works for you, I'd encourage you to keep using your fork :)

Thanks for your contribution.

PR #130, opened 2017-02-17
Added appearAnimation as a prop

deibeljc:


joshwcomeau:
Awesome! Thanks for the review.

I'll pull it and play with it over the weekend, just to check that there aren't unintended consequences. A quick read-through looks good to me though!

PR #134, opened 2017-03-31
Suppress warnings when in production

AlexanderOtavka:
Prop converter does several runtime checks and logs warnings that aren't needed in production.  When `process.env.NODE_ENV === 'production'`, they are now disabled.

### DIFF
--- a/src/prop-converter.js
+++ b/src/prop-converter.js
@@ -10,6 +10,7 @@
  *     be, if a single child is passed in.)
  *   - Resolving animation presets into their base CSS styles
  */
+/* eslint-disable block-scoped-var */
 
 import React, { Component, PropTypes } from 'react';
 
@@ -28,10 +29,13 @@ import {
 } from './enter-leave-presets';
 import { isElementAnSFC, omit } from './helpers';
 
-// Define process global in case the consumer hasn't defined process.env.NODE_ENV
+// Define `process` global in case the consumer hasn't defined
+// process.env.NODE_ENV
+/* eslint-disable no-use-before-define, vars-on-top, no-var */
 if (typeof process === 'undefined') {
   var process = { env: {} };  // Must use var to escape block scoping
 }
+/* eslint-enable */
 
 function propConverter(ComposedComponent) {
   class FlipMovePropConverter extends Component {
@@ -72,7 +76,7 @@ function propConverter(ComposedComponent) {
 
         if (isNaN(value)) {
           const defaultValue = defaultProps[prop];
-          
+
           if (process.env.NODE_ENV !== 'production') {
             console.error(invalidTypeForTimingProp({
               prop,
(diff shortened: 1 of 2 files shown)


joshwcomeau:
Thanks for your contribution!

Sorry for the delay on this; I'm interested in merging this PR, but I wanted to make a slight tweak, and it's been a hectic week (just started a new job!)

I'll leave a comment in the PR. If you can make the requested tweak I'll happy merge it in :)

src/prop-converter.js
```diff
@@ -41,16 +45,18 @@ function propConverter(ComposedComponent) {
...
+        // Check to see if any supplied components won't work.
+        // If the child doesn't have a key, it means we aren't animating it.
+        // It's allowed to be an SFC, since we ignore it.
+        const noStateless = workingProps.children.every(child =>
+           !isElementAnSFC(child) || typeof child.key === 'undefined'
+        );
+
+        if (!noStateless) {
+          console.warn(statelessFunctionalComponentSupplied());
+        }
```
joshwcomeau: Can we move this whole block into a method on the class? Something like:

```js
checkForStatelessFunctionalComponents() {
  if (process.env.NODE_ENV === 'production') {
    return;
  }

  // insert the chunk of original code here, with the comment, and the console.warn.
}
```

Then we'd invoke this method on like 48, so that the behaviour isn't changed (still runs at the same time)

src/prop-converter.js
```diff
@@ -28,6 +28,10 @@ import {
 } from './enter-leave-presets';
 import { isElementAnSFC, omit } from './helpers';
 
+// Define process global in case the consumer hasn't defined process.env.NODE_ENV
+if (typeof process === 'undefined') {
+  var process = { env: {} };  // Must use var to escape block scoping
+}
```
joshwcomeau: Hm. I _think_ instead we could do:

```js
// Get the 'global' scope. `window` in the browser, `global` on the server
const context = typeof window === 'undefined' ? global : window;

if (!context.process) {
  context.process = { env: {} };
}
```

AlexanderOtavka: The reason I used the local variable instead of setting the global is because I wanted to keep the module self-contained.  While I don't foresee any bugs around this, people do weird things in their code and I wanted to avoid modifying the global scope if at all possible.

src/prop-converter.js
AlexanderOtavka: Sure, will do.

src/prop-converter.js
joshwcomeau: Ahhh, yeah, great point. Leave it as-is :+1:

src/prop-converter.js
AlexanderOtavka: I tried it, but it's actually harder to figure out what is going on.  When it's inline, it's easy to see that the check doesn't happen in production and what is being checked.  Since we would only call this function in one place, it doesn't make much sense to make it its own function.

If you want me to do it anyway, I will.  It's your codebase.

joshwcomeau: Yeah, I just have a natural aversion to nested `if` statements, but you're probably right in that it's simpler to understand inline.

I may revise it later, but I'm happy to accept it as-is. I can tinker with it and see if it makes more sense.

PR #142, opened 2017-04-23
Refactor environment calculation

joshwcomeau:
Earlier this month, @AlexanderOtavka submitted a PR (#134) to strip console warnings from production.

The code works great, but it's a tricky thing to do well, since we don't know what environment the code will run in. It could be:

- Running in Node, where `process.env` is an object full of environment variables
- Running on the client, built by Webpack, where DefinePlugin overwrites `process.env.NODE_ENV` with a string like "production"
- Running on the client using the UMD build, where `process` is undefined and trying to access `process.env.NODE_ENV` throws an exception.

The existing solution defines a top-level variable to `process` and `process.env` if it doesn't exist, so that checks to `process.env.NODE_ENV` will never cause problems, but this strategy doesn't feel great to me.

This PR replaces it with a simpler try/catch. I was expecting a bit of a hit on file size (I assumed that Webpack would be able to dead-code-eliminate the `process.env.NODE_ENV` version), but after checking the file sizes, this PR actually comes in slightly _smaller_; master is currently 6.3k minified and gzipped, but this PR goes down to 5.8k. 

(I'm not 100% sure why this is. Looking at the code, master has a bunch of stuff related to intervals that this PR is missing. Gonna dig into this a bit. But either way, there isn't any significant dead code elimination either way).

@AlexanderOtavka could you please review? I want to ensure that I haven't inadvertently broken something (this feels a little too good to be true). My tests with Storybook seem to work fine regardless of environment, but a sanity check would be appreciated :)

@rhys-vdw tagged you as well in case you have any insights :)

### DIFF
--- a/src/prop-converter.js
+++ b/src/prop-converter.js
@@ -30,17 +30,35 @@ import {
 import { isElementAnSFC, omit } from './helpers';
 
 
-// Define `process` global in case the consumer hasn't defined
-// process.env.NODE_ENV. We use 'var' to escape block scoping.
-/* eslint-disable no-use-before-define, vars-on-top, no-var */
-if (typeof process === 'undefined') {
-  var process = { env: {} };
+let nodeEnv;
+try {
+  nodeEnv = process.env.NODE_ENV;
+} catch (e) {
+  nodeEnv = 'development';
 }
-/* eslint-enable */
-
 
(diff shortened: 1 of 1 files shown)


AlexanderOtavka:
Working fine for me.

PR #144, opened 2017-05-01
Remove direct usage of opaque children object

AlexanderOtavka:
Per the react spec, `props.children` should be treated as an opaque object, not an array.  The `React.Children` api should be used to handle this object.

The current implementation was working fine with regular react, but these changes are necessary for compatibility with substitutes like preact.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -7,7 +7,7 @@
 
  /* eslint-disable react/prop-types */
 
-import React, { Component } from 'react';
+import React, { Component, Children } from 'react';
 
 import './polyfills';
 import propConverter from './prop-converter';
@@ -69,7 +69,7 @@ class FlipMove extends Component {
     // can complete. Because we cannot mutate props, we make `state` the source
     // of truth.
     this.state = {
-      children: props.children.map(child => ({
+      children: Children.toArray(props.children).map(child => ({
         ...child,
         appearing: true,
       })),
@@ -108,17 +108,20 @@ class FlipMove extends Component {
     // so it can be used later to work out the animation.
     this.updateBoundingBoxCaches();
 
+    // Convert opaque children object to array.
+    const nextChildren = Children.toArray(nextProps.children);
+
     // Next, we need to update our state, so that it contains our new set of
     // children. If animation is disabled or unsupported, this is easy;
     // we just copy our props into state.
     // Assuming that we can animate, though, we have to do some work.
     // Essentially, we want to keep just-deleted nodes in the DOM for a bit
     // longer, so that we can animate them away.
-    const newChildren = this.isAnimationDisabled(nextProps)
-      ? nextProps.children
-      : this.calculateNextSetOfChildren(nextProps.children);
-
-    this.setState({ children: newChildren });
+    this.setState({
+      children: this.isAnimationDisabled(nextProps)
+        ? nextChildren
+        : this.calculateNextSetOfChildren(nextChildren)
+    });
   }
 
   componentDidUpdate(previousProps) {
@@ -129,8 +132,8 @@ class FlipMove extends Component {
     // At the end of the transition, we clean up nodes that need to be removed.
     // We DON'T want this cleanup to trigger another update.
 
-    const oldChildrenKeys = this.props.children.map(d => d.key);
-    const nextChildrenKeys = previousProps.children.map(d => d.key);
+    const oldChildrenKeys = Children.toArray(this.props.children).map(d => d.key);
+    const nextChildrenKeys = Children.toArray(previousProps.children).map(d => d.key);
 
     const shouldTriggerFLIP = (
       !arraysEqual(oldChildrenKeys, nextChildrenKeys) &&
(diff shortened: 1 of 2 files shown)


src/FlipMove.js
```diff
@@ -69,7 +69,7 @@ class FlipMove extends Component {
     // can complete. Because we cannot mutate props, we make `state` the source
     // of truth.
     this.state = {
-      children: props.children.map(child => ({
+      children: Children.toArray(props.children).map(child => ({
```
joshwcomeau: Since `prop-converter` handles converting `children` to array, is it necessary in this file?

The goal of `prop-converter` is to abstract away all the tedious business of stuff like this, so if we can isolate it to that HOC I'd be happy :)

That said, I don't have much experiencing aliasing React-like libraries, so let me know if there's a reason that this isn't doable.

AlexanderOtavka: The problem is that `propConverter` composes around the `FlipMove` component.  That means the children object passes back through the React virtual DOM and so the children prop passed to `FlipMove` may not be the same one that came from `propConverter`.  Unfortunately, we have to do all of the children conversions in `FlipMove` instead of `propConverter` because `children` is a special prop.

joshwcomeau: Hm.. The thing is, it shouldn't be possible to access `FlipMove` without going through `propConverter`. The export of `FlipMove.js` is `propConverter(FlipMove)`, so when someone imports it, they're actually importing the `FlipMovePropConverter` component.

Is there an internal mechanism with children I'm not familiar with?

AlexanderOtavka: Yes, but in the `render` method of the prop converter, it renders a new `ComposedComponent` which is a `FlipMove` (as it should).  That means the converted props get passed back into the React virtual dom, which may modify the `children` prop before it gets to the `FlipMove` class.  Thus, a HOC cannot handle conversion of this particular prop; it must be done in `FlipMove`.

joshwcomeau: Sorry, still not sure what you mean by "passed back into the React virtual dom". Where would the modifications be coming from?

Let's say we have the following setup:

```jsx
class Whatever extends Component {
    render() {
        return (
            <FlipMove>
                <div key={1} />
                <div key={2} />
            </FlipMove>
        );
    }
}
```

If we inspect the React tree, it would look like this:

```jsx
<FlipMovePropConverter>
    <ComposedComponent> // AKA Flip Move
        <div key={1} />
        <div key={2} />
    </ComposedComponent>
</FlipMovePropConverter>
```

Let's say the parent component `Whatever` updates, so that two children are removed and a third is added:

```jsx
class Whatever extends Component {
    render() {
        return (
            <FlipMove>
                <div key={3} />
            </FlipMove>
        );
    }
}
```

`FlipMovePropConverter` will re-render, creating a new React element that has this new `children` prop of a single object. It's still going to run through the same process, though; on every `render`, `FlipMovePropConverter` calls `this.convertProps`, which will do the React.Children work.

Apologies for the back-and-forth, but I'd like to understand what you mean. As I understand it, React works by creating new elements on every render, and reconciling the differences between each virtual tree, doing DOM updates when the diff shows that a change is needed. I don't see how this process would undo the work of the prop converter, so if there _is_ some missing piece I'm not aware of, I'd like to learn about it :)

AlexanderOtavka: In `flip-move.js`, we have the following:

```jsx
render() {
  return (
    <ComposedComponent {...this.convertProps(this.props)} />
  );
}
```

But that is really the same as doing this:

```jsx
render() {
  return React.createElement(ComposedComponent, this.convertProps(this.props));
}
```

So the props are passed back to React, which handles calling the constructor and lifecycle callbacks of `ComposedComponent`.  React may not pass the same props object to `ComposedComponent` that was passed to `React.createElement`, in particular, it is free to change the children object from an array to whatever.  That is what preact does, hence the issue.

That means we need to do the conversion of children to an array in `FlipMove`.

joshwcomeau: Ahh I see, so the issue is that React-like libraries may mutate the supplied props in-between layers. Even though we supply an array as the `children` prop, libraries internally munge it into a format that works for them.

Thanks for the clarification.

src/FlipMove.js
```diff
@@ -238,7 +241,7 @@ class FlipMove extends Component {
     // we need to reset the transition, so that the NEW shuffle starts from
     // the right place.
     this.state.children.forEach((child) => {
-      const { domNode } = this.childrenData[child.key];
+      const { domNode } = this.childrenData[child.key] || {};
```
joshwcomeau: Was this to fix an issue with Preact, or just generally a good idea to avoid issues?

AlexanderOtavka: Yes, there was another issue with Preact, I think it resolves ref callbacks a bit later than React does, but this seemed to fix the issue.

AlexanderOtavka:
Is this ready to be merged?

joshwcomeau:
Sorry for the delay - I wanted to quickly run the stories myself, and it's been a hectic week.

Thank you for your contribution :) I'm sure Preact/Inferno users will be delighted that it's compatible.

joshwcomeau:
Published in `2.9.4`. I've also sent an invitation to add you as a collaborator on Flip Move :)

The idea is that for those like you who have made high-quality contributions, you should have more control over this project. Feel free to review PRs submitted by others, close solved issues, merge stuff that has at least 1 other approval, etc. 

Of course it's up to you if you want to do any of this, but the option is there if it interests you!

AlexanderOtavka:
Awesome, thanks!  I'll do what I can.

Issue #139, opened 2017-04-18
Replace PropTypes with Flow


joshwcomeau:
I've been using Flow on other projects, and I like it.

If anyone wants to lead this change, lemme know. Otherwise I'll likely tackle it in the next couple weeks.

Hypnosphi:
I think I can do that

Hypnosphi:
OK, there is a problem.

[Here](https://github.com/joshwcomeau/react-flip-move/blob/master/src/prop-converter.js#L127) `propTypes` keys are used in runtime to determine which props should be delegated to the DOM element. This isn't really achievable with static typechecking.

Would it be OK to keep an array of FlipMove-specific prop keys in a separate variable and keep it in sync with props typings?

Hypnosphi:
OK, I think I found a better solution: https://github.com/joshwcomeau/react-flip-move/pull/145/files#diff-ecdb6cdeb1de0c9f5925708a2be17e73R122

joshwcomeau:
Right, yeah that's tricky.

I like your solution, though! 

I'm wondering if we ought to keep some PropTypes as well, as a helpful indicator to developers using FlipMove. Flow is great for contributors to FlipMove, but it offers no runtime notifications for developers who may be using FlipMove incorrectly.

I don't think it's that big a deal - we already log warnings when developers do things they shouldn't (eg. using SFCs). Just wanted to bring it up in case anyone else had opinions :)

Hypnosphi:
I'll make another little PR which would just replace `React.PropTypes` with `prop-types` package then

PR #145, opened 2017-05-03
Use Flow instead of React.PropTypes

### LINKED ISSUE #138: React 15.5 PropType deprecation warnings

Hypnosphi:
fixes #138 & #139


test/index.spec.js
```diff
@@ -228,21 +228,21 @@ describe('FlipMove', () => {
...
       expect(consoleStub).to.have.been.calledOnce;
     });
 
     it('applies an array prop and throws', () => {
-      renderedComponent.setState({ duration: [10] });
+      renderedComponent.setState({ duration: ['hi'] });
       expect(consoleStub).to.have.been.called;
     });
+
+    it('applies a string that can be converted to an int', () => {
```
Hypnosphi: I needed to move the "correct" testcase down, because all these tests persist state between cases, and some of them rely on that. This is not good in fact, but it's a separate issue.

Hypnosphi:
Looks like this shouldn't yet be merged. Flow types are lost if the typed value goes through untyped files, so `FlipMove.js` also has to be typed

test/index.spec.js
joshwcomeau: Yeah, the tests in this project are kind of a disaster. Thanks for wading through them! It's on my list to improve this stuff.

Hypnosphi:
>Let me know when FlipMove is typed

Here it is

Hypnosphi:
Now it's really ready for review

package.json
```diff
@@ -10,11 +10,14 @@
...
-    "build:lib": "babel src --out-dir lib",
+    "build:lib": "babel src --out-dir lib && flow-copy-source -v src lib",
     "build:umd": "webpack src/index.js dist/react-flip-move.js --config webpack.config.production.js",
     "build:umd:min": "webpack src/index.js dist/react-flip-move.min.js --config webpack.config.production.min.js",
     "build": "npm run build:lib && npm run build:umd && npm run build:umd:min",
-    "prepublish": "npm run build",
+    "flow": "flow",
+    "flow-coverage": "flow-coverage-report -i 'src/**/*.js' -t html -t json -t text",
+    "prebuild": "npm run lint && flow check && npm run test",
+    "prepublishOnly": "npm run build",
```
joshwcomeau: Yay, switch to `prepublishOnly`.

I'll just need to be careful I don't try publishing on a machine running NPM <4, but I think that's reasonable in exchange for cutting out the annoying build-on-install.

package.json
```diff
@@ -10,11 +10,14 @@
...
     "lint": "eslint src",
-    "build:lib": "babel src --out-dir lib",
+    "build:lib": "babel src --out-dir lib && flow-copy-source -v src lib",
     "build:umd": "webpack src/index.js dist/react-flip-move.js --config webpack.config.production.js",
     "build:umd:min": "webpack src/index.js dist/react-flip-move.min.js --config webpack.config.production.min.js",
     "build": "npm run build:lib && npm run build:umd && npm run build:umd:min",
-    "prepublish": "npm run build",
+    "flow": "flow",
+    "flow-coverage": "flow-coverage-report -i 'src/**/*.js' -t html -t json -t text",
+    "prebuild": "npm run lint && flow check && npm run test",
```
joshwcomeau: Neat. I wonder if there's a way to run this before `build`, and to abort if any of the steps fail?

If so, that would be great. Would prevent publishing a build with lint/flow errors or test failures.

package.json
```diff
@@ -10,11 +10,14 @@
   ],
   "scripts": {
     "lint": "eslint src",
-    "build:lib": "babel src --out-dir lib",
+    "build:lib": "babel src --out-dir lib && flow-copy-source -v src lib",
     "build:umd": "webpack src/index.js dist/react-flip-move.js --config webpack.config.production.js",
     "build:umd:min": "webpack src/index.js dist/react-flip-move.min.js --config webpack.config.production.min.js",
```
joshwcomeau: Is anything needed for the UMD builds? I imagine we'd still have to remove all the types somehow.

src/dom-manipulation.js
```diff
@@ -7,20 +8,32 @@
  * only spot in the codebase with impure functions.
  */
 import { findDOMNode } from 'react-dom';
-
-
-export function applyStylesToDOMNode({ domNode, styles }) {
+import type { Component } from 'react';
```
joshwcomeau: Woah, I didn't know this was a thing!

src/dom-manipulation.js
```diff
@@ -7,20 +8,32 @@
...
+  styles: Styles,
+}) {
   // Can't just do an object merge because domNode.styles is no regular object.
   // Need to do it this way for the engine to fire its `set` listeners.
   Object.keys(styles).forEach((key) => {
-    // eslint-disable-next-line no-param-reassign
-    domNode.style[key] = styles[key];
+    const value = styles[key];
+    const stringValue = typeof value === 'number' ? `${value}px` : value;
+    domNode.style.setProperty(key, stringValue);
```
joshwcomeau: Hah, you ran into this guy: https://github.com/facebook/flow/issues/3733

(I started working on this issue myself, but I didn't get much farther than this).

I like your solution, but I see a small issue with it: the style being set might not have a pixel value. For example, `opacity` is something often used in enter/leave animations.

Could we not just do:

```js
const value = styles[key];
domNode.style.setProperty(key, String(value));
```

Off the top of my head, I can't think of any reason that this wouldn't work, but it should definitely be tested with the included stories (and maybe even adding one that defines a custom enterAnimation that animates `height` or something equally heinous)

src/dom-manipulation.js
```diff
@@ -35,7 +48,7 @@ export function whichTransitionEvent() {
   const el = document.createElement('fakeelement');
 
   const match = Object.keys(transitions).find(t => (
-    el.style[t] !== undefined
+    el.style.getPropertyValue(t) !== undefined
```
joshwcomeau: :+1:

.eslintrc
```diff
@@ -1,20 +1,28 @@
...
+    "flowtype/no-dupe-keys": 2,
+    "flowtype/no-primitive-constructor-types": 2,
+    "flowtype/no-weak-types": 1,
+    "flowtype/object-type-delimiter": [2, "comma"],
+    "flowtype/semi": 2,
+  },
+  settings: {
+    flowtype: {
+      onlyFilesWithFlowAnnotation: true,
+    }
```
joshwcomeau: Could you elaborate on why this `settings` object is required? I was under the impression that only files with `// @flow` would be parsed.

.travis.yml
```diff
@@ -9,4 +9,6 @@ before_script:
   - export CHROME_BIN=chromium-browser
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
-  - npm install babel-cli -g
+  - yarn global add babel-cli
+script:
+  - yarn build
```
joshwcomeau: I've had issues using yarn with CI before. Maybe they've been straightened out by now... but I think I'd rather go back to NPM. 

(perhaps I'm being overly cautious, but I once lost a day of work tracing down an incomprehensible CI error message that was because of a bug in Yarn. Flip Move runs pretty quick on CI anyway)

src/dom-manipulation.js
```diff
@@ -68,36 +82,36 @@ export const getRelativeBoundingBox = ({
 
 /** getPositionDelta
  * This method returns the delta between two bounding boxes, to figure out
- * how mant pixels on each axis the element has moved.
+ * how many pixels on each axis the element has moved.
```
joshwcomeau: Thanks for fixing the typo :)

.eslintrc
Hypnosphi: It overrides the setting from recommended config: https://github.com/gajus/eslint-plugin-flowtype/blob/master/src/configs/recommended.json

package.json
Hypnosphi: Not really, it's for Flow users who'll import the library via CommonJS or ES modules. It just copies all the source files as is, appending `.flow` to the filename

package.json
Hypnosphi: npm does exactly that automatically =)

src/dom-manipulation.js
Hypnosphi: Sounds reasonable, will revert

Hypnosphi:
>hopefully it isn't too much trouble to leave those private variables exposed

It will require some weird typing like `$Shape<Element<*>> & {...}`, but it's doable. I would keep the typing of hooks though

.travis.yml
Hypnosphi: OK

Hypnosphi:
I had to introduce a new `ElementShape` type, because `$Shape` didn't seem to work well with classes. So i had to keep the reference to real element in `ChildData`, so that in could be passed to `React.cloneElement`

src/helpers.js
```diff
@@ -8,17 +11,17 @@ export const isElementAnSFC = (element) => {
   return !element.type.prototype.isReactComponent;
 };
 
-export function omit(obj, attrs = []) {
-  const result = {};
-  Object.keys(obj).forEach((key) => {
+export function omit<R: {}, T: R>(obj: T, attrs: $Keys<T>[] = []): R {
+  const result: $Shape<T> = {};
+  Object.keys(obj).forEach((key: $Keys<T>) => {
```
joshwcomeau: For fun, a friend and I were trying to see if we could come up with a way within Flow to specify `omit`'s behaviour, using `$Diff` and `$Keys`. We couldn't quite get there, alas, but my friend did come up with what I think is a simpler implementation:

```js
export function omit<T: {}>(obj: T, attrs: Array<$Keys<T>> = []): $Shape<T> {
  const result = {};
  Object.keys(obj).forEach((key: $Keys<T>) => {
    if (attrs.indexOf(key) === -1) {
      result[key] = obj[key];
    }
  });
  return result;
}
``` 

Let me know if there's a reason you need the second generic.

src/FlipMove.js
```diff
@@ -180,15 +194,15 @@ class FlipMove extends Component {
     // into the nextChildren in its ORIGINAL position. Note that, as we keep
     // inserting old items into the new list, the "original" position will
     // keep incrementing.
-    let numOfChildrenLeaving = 0;
-    this.state.children.forEach((child, index) => {
-      const isLeaving = !nextChildren.find(({ key }) => key === child.key);
+    let numOfChildrenLeaving: number = 0;
```
joshwcomeau: This seems like an annotation that Flow could infer, unless there's a reason that you wanted to be explicit here?

src/FlipMove.js
```diff
@@ -397,49 +416,55 @@ class FlipMove extends Component {
...
+      if (this.heightPlaceholderData.domNode) {
+        this.heightPlaceholderData.domNode.style.height = '0';
       }
     }
   }
 
-  formatChildrenForHooks() {
-    const elements = [];
-    const domNodes = [];
+  callChildrenHook(hook: ChildrenHook) {
```
joshwcomeau: I like this change in abstraction :+1:

src/FlipMove.js
```diff
@@ -397,49 +416,55 @@ class FlipMove extends Component {
...
+    const elements: ElementShape[] = [];
+    const domNodes: (?HTMLElement)[] = [];
```
joshwcomeau: Can we use the longform `Array<>` form when maybe types are involved? Feels cleaner to me.

(I was actually going to ask that we convert to it everywhere, as it's the format preferred by Facebook internally to avoid exactly this kind of ambiguity, but I didn't wanna add more work to your plate. If you're energetic enough, I'd appreciate it, but it's by no means necessary).

src/FlipMove.js
```diff
@@ -450,32 +475,27 @@ class FlipMove extends Component {
...
-  computeInitialStyles(child) {
-    const noAnimationRequestedForThisEvent = (
-      (child.appearing && !this.props.appearAnimation) ||
-      (child.entering && !this.props.enterAnimation) ||
-      (child.leaving && !this.props.leaveAnimation)
-    );
-
-    if (noAnimationRequestedForThisEvent) {
-      return {};
-    }
```
joshwcomeau: Why'd you kill this? IMO it's better to do calculations like this upfront, to simplify the conditionals further down.

src/FlipMove.js
```diff
@@ -119,21 +94,21 @@ class FlipMove extends Component {
     // longer, so that we can animate them away.
     this.setState({
       children: this.isAnimationDisabled(nextProps)
-        ? nextChildren
-        : this.calculateNextSetOfChildren(nextChildren)
+        ? nextChildren.map(element => ({ ...element, element }))
```
joshwcomeau: Curious why this was changed? I know earlier you spoke about keeping this type of object:

```js
{ element: ReactElement, appearing: boolean, leaving: boolean, entering: boolean }
```

The way this is now, though, we wind up nesting the element within itself:

```js
{
  element: ReactElement,
  ...ReactElement,
  appearing: boolean,
  leaving: boolean,
  entering: boolean,
}
```

Curious if this is intentional, or a "leftover" from the earlier change.

src/dom-manipulation.js
```diff
@@ -214,10 +234,17 @@ export const getNativeNode = (element) => {
...
   // composite components.
-  return findDOMNode(element);
+  const foundNode: ?(Element | Text) = findDOMNode(element);
+
+  if (foundNode instanceof HTMLElement) {
+    return foundNode;
+  }
+
+  // Text nodes are not supported
+  return null;
```
joshwcomeau: Good catch.

Nitpick: I find it clearer when invalid cases are dealt with before the expected case, so that the last line in a function is the desired result. I would reorder:

```js
const isHTMLElement = foundNode instanceof HTMLElement;

if (!isHTMLElement) {
  // Text nodes are not supported
  return null;
}

return foundNode;
```

src/prop-converter.js
```diff
@@ -29,18 +33,41 @@ import {
...
+  Presets,
+  FlipMoveProps,
+  FlipMoveDefaultProps,
+  ConvertedProps,
+  DelegatedProps,
+} from './typings';
+
+const nodeEnv: string = process && process.env && process.env.NODE_ENV
+  ? process.env.NODE_ENV
+  : 'development';
```
joshwcomeau: Why did this change?

package.json
```diff
@@ -69,15 +78,12 @@
     "karma-webpack": "1.7.0",
     "lodash": "4.3.0",
     "mocha": "2.4.5",
+    "prop-types": "15.5.8",
```
joshwcomeau: After you rebase, please remove this from `devDependencies` :)

src/typings.js
```diff
@@ -0,0 +1,117 @@
+// @flow
+import type { Element } from 'react';
+
+export type Styles = {
+  [string]: string,
+};
+
+type ReactStyles = {
+  [string]: string | number,
+};
```
Hypnosphi: It can be a bit confusing, because when you supply `height: 20`, we don't append `px` (but people may expect us to do it, because React does), and height doesn't get set to desired value

src/prop-converter.js
Hypnosphi: Flow points to possible runtime errors even if they happen inside `try` block

src/prop-converter.js
```diff
@@ -28,18 +33,41 @@ import {
...
-  nodeEnv = process.env.NODE_ENV;
-} catch (e) {
-  nodeEnv = 'development';
-}
-
-function propConverter(ComposedComponent) {
-  class FlipMovePropConverter extends Component {
-    checkForStatelessFunctionalComponents(children) {
+    // eslint-disable-next-line class-methods-use-this
+    checkForStatelessFunctionalComponents(children: mixed) {
```
Hypnosphi: Exactly. `mixed` means that the value is opaque and you can make no assumptions about it

src/FlipMove.js
```diff
@@ -22,70 +27,40 @@ import {
...
+  DelegatedProps,
+  Styles,
+} from './typings';
 
 const transitionEnd = whichTransitionEvent();
 const noBrowserSupport = !transitionEnd;
 
+function getKey(childData: ChildData): string {
+  return childData.key || '';
+}
```
Hypnosphi: This was more helpful when it was `childData.element.key || ''`, but I decided to keep it, because when the shape of `ChildData` changes again, it will be a matter of changing the helpe's body instead of ten replacements

src/FlipMove.js
```diff
@@ -22,70 +27,40 @@ import {
...
+  // receives a new set of props that do NOT contain the item.
+  // If we just render the props as-is, the item would instantly disappear.
+  // We want to keep the item rendered for a little while, until its animation
+  // can complete. Because we cannot mutate props, we make `state` the source
+  // of truth.
+  state = {
+    children: Children.toArray(this.props.children).map((element: Element<*>) => ({
+      ...element,
+      element,
+      appearing: 3,
```
Hypnosphi: Definitely a typo

src/FlipMove.js
Hypnosphi: We need to keep the reference to real `ReactElement`, because we pass it to `createElement`

src/FlipMove.js
Hypnosphi: I'm ok with using explicit `Array<>` syntax everywhere

src/helpers.js
Hypnosphi: The second jeneric lets you specify exact type you want to return: `const B: Type = omit(A, ...keys)`

Hypnosphi:
>I've been pleased that through a few improvements, the size has remained under 6kb.

Can you please commit the improvements you made, I'll try to investigate further

src/FlipMove.js
Hypnosphi: Right, this can be omitted safely

src/FlipMove.js
Hypnosphi: I didn' kill it, just splitted and moved below. Flow isn't clever enough to find out that if `noAnimation...` is falsy, and `child.appearing` is truthy,  `this.props.appearAnimation` cannot be null

package.json
Hypnosphi: Why? It stays used only in stories

src/index.js
```diff
@@ -1,5 +1,8 @@
+// @flow
 /**
  * React Flip Move
  * (c) 2016-present Joshua Comeau
  */
-module.exports = require('./FlipMove');
+import FlipMove from './FlipMove';
+
+export default FlipMove;
```
Hypnosphi: `export default from '...'` is on [stage 1](https://github.com/leebyron/ecmascript-export-default-from) yet, and Flow doesn't understand it.

And Flow loses type information when you use commonJs imports with `require`.

This:
```js
import FlipMove from './FlipMove'
module.exports = FlipMove
```
seems to work though, but it shouldn't be really needed, because you use `babel-plugin-add-module-exports`anyway, and `export default FlipMove` gets compiled to
```js
exports.default = _FlipMove2.default;
module.exports = exports['default'];
```
given that there's no named exports.

src/prop-converter.js
joshwcomeau: Hm, I think we should just tell Flow not to worry about it with `$FlowFixMe` then. 

The problem with this solution is that webpack's `definePlugin` will transform `process.env.NODE_ENV`. So it'll transform into this:

```js
const nodeEnv: string = process && process.env && 'production'
  ? 'production'
  : 'development';
```

Which will erroneously resolve to 'development'.

joshwcomeau:
> Can you please commit the improvements you made, I'll try to investigate further

Sorry, what I mean is that over the past ~6 months, Flip Move has added more functionality without significantly increasing its gzip filesize. So, if possible I'd like to continue that trend.

src/prop-converter.js
Hypnosphi: Oh I see

Hypnosphi: Good news: after fixing that, gzip size is 5866 bytes

src/FlipMove.js
```diff
@@ -575,38 +552,99 @@ class FlipMove extends Component {
...
+    const shouldTriggerFLIP = (
+      !arraysEqual(oldChildrenKeys, nextChildrenKeys) &&
+      !this.isAnimationDisabled(this.props)
+    );
+
+    if (shouldTriggerFLIP) {
+      this.prepForAnimation();
+      this.runAnimation();
+    }
+  }
```
joshwcomeau: Ah, sorry, I was unclear. I meant for the lifecycle methods to still be above the other custom methods, but below the initialized data (the stuff that used to be in the constructor). So they should sit just above `calculateNextSetOfChildren`.

src/index.js
joshwcomeau: Good call on skipping `export default from`, didn't realize it wasn't a part of the initial module syntax.

And yeah if it seems to work, I'm alright with it. I remember that there was a bug with something, but it's been a year, and a bunch of other libraries use the same build system without issues, so -shrugs-.

src/prop-converter.js
```diff
@@ -29,18 +33,50 @@ import {
...
+  FlipMoveDefaultProps,
+  ConvertedProps,
+  DelegatedProps,
+} from './typings';
+
+declare var process: {
+  env: {
+    NODE_ENV: 'production' | 'development',
+  },
+};
```
joshwcomeau: Good solution :+1:

src/FlipMove.js
Hypnosphi: Unfortunately, `react/sort-comp` [doesn't](https://github.com/yannickcr/eslint-plugin-react/issues/599) offer a way to separate properties from methods. One way to solve this is to stop using property initializers, but in that case we'd have to split property typings from values, like this:
```js
foo: FooType;
bar: BarType;
baz: BazType;

constructor(...args) {
  super(...args);

  this.foo = fooValue;
  this.bar = barValue;
  this.baz = bazValue;
}
```

Another option is to drop `react/sort-comp`, and use [`sort-class-members`](https://www.npmjs.com/package/eslint-plugin-sort-class-members) instead, which is more flexible

Hypnosphi:
I just realized that `setProperty` requires the name to be hyphenated, so for example `transformOrigin` wasn't set and accordion animations didn't look as they should.

src/helpers.js
```diff
@@ -33,3 +36,17 @@ export function arraysEqual(a, b) {
...
+  return (str) => {
+    if (!cache[str]) {
+      cache[str] = fn(str);
+    }
+    return cache[str];
+  };
+}
+
+export const hyphenate = memoizeString(str =>
+  str.replace(/([A-Z])/g, '-$1').toLowerCase());
```
joshwcomeau: Clever!

src/helpers.js
```diff
@@ -33,3 +36,17 @@ export function arraysEqual(a, b) {
 
   return a.every((element, index) => element === b[index]);
 }
+
+function memoizeString<T>(fn: (string) => T): (string) => T {
```
joshwcomeau: I found this type signature really hard to parse. Is there a way to pull the `(string) => T` part into a Type?

This is invalid, but I'm wondering if anything exists similar to:

```js
type StringManipulator = (string) => T;

function memoizeString<T>(fn: StringManipulator<T>): StringManipulator<T> 
```

src/FlipMove.js
joshwcomeau: Right. Thanks for the info, I'll check out `sort-class-members`

src/helpers.js
Hypnosphi: StringManipulator should also be generic. I think I'll do something like this:

```js
// 1 stands for 1 argument
type function1<I, O> = (I) => O
type higherOrderFunction1<I, O> = (function1<I, O>) => function1<I, O>
const memoizeString<T>: higherOrderFunction1<string, T> = ...
```

maybe with `HOF1` instead of `higherOrderFunction1`

Hypnosphi:
What do you thing about class members sorting options? https://github.com/joshwcomeau/react-flip-move/pull/145#discussion_r115543385

joshwcomeau:
For sorting, I'd like to try `sort-class-members`. I'm not super concerned about it, though; given that this project only has 2 components, I'm fine to just disable the rule if `sort-class-members` isn't nice and simple.

Random question maybe you can help with: After pulling these changes into my local repo and installing dependencies, I get the following error when I run `yarn flow`:

```
examples/node_modules/fbjs/lib/PromiseMap.js.flow:15
 15: var Deferred = require('Deferred');
                    ^^^^^^^^^^^^^^^^^^^ Deferred. Required module not found

examples/node_modules/fbjs/lib/fetchWithRetries.js.flow:16
 16: var ExecutionEnvironment = require('ExecutionEnvironment');
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ExecutionEnvironment. Required module not found

examples/node_modules/fbjs/lib/fetchWithRetries.js.flow:19
 19: var sprintf = require('sprintf');
                   ^^^^^^^^^^^^^^^^^^ sprintf. Required module not found

examples/node_modules/fbjs/lib/fetchWithRetries.js.flow:20
 20: var fetch = require('fetch');
                 ^^^^^^^^^^^^^^^^ fetch. Required module not found
```

I tried cloning a fresh copy and everything works, so it's likely something with my local config? Deleting `node_modules` and reinstalling didn't help.

Not a huge deal since I can just delete and re-clone it, but curious what it's about if you have any idea :)

joshwcomeau:
If you wanted to open a new PR for `sort-class-members` that'd be awesome, but you shouldn't feel obligated; you've already done a ton of work, I'm happy to sort out (heh) this last concern.

Hypnosphi:
>curious what it's about if you have any idea :)

No idea, sorry

joshwcomeau:
Alright, this and #150 are released in `2.9.9-beta1`. Thanks again for your efforts on this!

I've invited to you to become a collaborator on this project :)

The idea is that for those like you who have made high-quality contributions, you should have more control over this project. Feel free to review PRs submitted by others, close solved issues, merge stuff that has at least 1 other approval, etc.

Of course it's up to you if you want to do any of this, but the option is there if it interests you!

PR #146, opened 2017-05-03
Replace `React.PropTypes` with `prop-types` package

### LINKED ISSUE #138: React 15.5 PropType deprecation warnings

Hypnosphi:
Fixes #138

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "mocha": "2.4.5",
     "react": "15.5.4",
     "react-addons-test-utils": "15.0.0",
-    "react-create-class": "1.0.0",
     "react-dom": "15.5.4",
     "sinon": "1.17.3",
     "sinon-chai": "2.8.0",
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "babel-preset-react": "6.3.13",
     "babel-preset-stage-0": "6.1.18",
     "chai": "3.5.0",
+    "create-react-class": "15.5.2",
     "css-loader": "0.23.1",
     "eslint": "3.10.0",
     "eslint-config-airbnb": "12.0.0",
@@ -68,12 +69,16 @@
     "karma-webpack": "1.7.0",
     "lodash": "4.3.0",
     "mocha": "2.4.5",
-    "react": "15.2.1",
+    "react": "15.5.4",
     "react-addons-test-utils": "15.0.0",
-    "react-dom": "15.0.0",
+    "react-create-class": "1.0.0",
+    "react-dom": "15.5.4",
     "sinon": "1.17.3",
     "sinon-chai": "2.8.0",
     "style-loader": "0.13.1",
     "webpack": "1.12.11"
+  },
+  "dependencies": {
+    "prop-types": "15.5.8"
   }
 }
--- a/src/prop-converter.js
+++ b/src/prop-converter.js
@@ -12,7 +12,8 @@
  */
 /* eslint-disable block-scoped-var */
 
-import React, { Component, PropTypes, Children } from 'react';
+import React, { Component, Children } from 'react';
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -68,12 +69,16 @@
     "karma-webpack": "1.7.0",
     "lodash": "4.3.0",
     "mocha": "2.4.5",
-    "react": "15.2.1",
+    "react": "15.5.4",
     "react-addons-test-utils": "15.0.0",
-    "react-dom": "15.0.0",
+    "react-create-class": "1.0.0",
```
joshwcomeau: Looks like this package isn't necessary (apparently they switched names to `create-react-class`, which you have above)

Hypnosphi: Good catch, fixed that

Hypnosphi:
>but we'll be able to remove this package (or, at least move it to devDependencies, since we may still want it in stories) when we merge in your Flow PR

It depends on whether we decide to keep prop-types to help the users that don't have Flow, as you've suggested in https://github.com/joshwcomeau/react-flip-move/issues/139#issuecomment-298946748

joshwcomeau:
> It depends on whether we decide to keep prop-types to help the users that don't have Flow, as you've suggested in #139 (comment)

Ohh, right. Sorry about that! Didn't mean to contradict myself.

Thinking about the issues I've seen recently, the mistakes people make are often things not caught by PropTypes. For example, #147 was because of an `easing` that doesn't exist in CSS (and we can't use `PropTypes.oneOf` since `cubic-bezier` has values).

I think we should abolish PropTypes checks, but replace them with a few additional development-only runtime checks for stuff like that.

I can add some quick checks after I merge your two branches.

Hypnosphi:
But the only purpose of this PR was to have a release without warnings while #145 is being reviewed. There's no sense in merging both simultaneously

joshwcomeau:
> But the only purpose of this PR was to have a release without warnings while #145 is being reviewed. There's no sense in merging both simultaneously

Sure there is! This release also bumps the development react version, switches the dependency for createClass, uses the prop-types dependency for internal components (for stories), updates the test utils...

If the `prop-types` thing was urgent, I'd release it as a stop-gap until the Flow one is ready. In my opinion, though, the churn of introducing something and then removing it within a few days is a bigger issue than a developer console warning. And I should be able to finish reviewing the Flow branch within the next couple days, so it shouldn't be a long wait period.

Sorry again for being unclear about whether I wanted to keep prop-types in the build. This is still a worthwhile contribution though :)

Hypnosphi:
>This release also bumps the development react version, switches the dependency for createClass, <...>, updates the test utils...

This is also done in #145. Only 'prop-types' in stories are missing.

Those warnings are the only thing that blocks my project from trying `react@next`. And I don't think that such an in-place replacement is really introducing something new

joshwcomeau:
Yeah, I mean I was actually hoping to get the Flow changes merged in this weekend as well, making the time interval negligible. Looking at it more closely, though, it's going to take me a while to get through that review (in no small part because I need to learn a lot about Flow as I go, lol).

Gonna push this through now, and will hopefully still submit a review by tomorrow.

joshwcomeau:
Published in `v2.9.6`.

Let me know how it goes with trying React 16 :)

Hypnosphi:
>I realized right after publish that it made more sense to move this dependency to peerDependencies, and to loosen the version restriction to 15.x.x. This way, users can bring their own version, and NPM won't complain if theirs is a patch or minor version off.

There are bugs in versions prior to `15.5.7`. The recommended way for libraries is to have it in `dependencies` with `^` prepended: https://github.com/reactjs/prop-types#how-to-depend-on-this-package

joshwcomeau:
Argh, of course, lol. Publishing yet another version now.

PR #151, opened 2017-05-13
Fix preact incompatibilities

AlexanderOtavka:
Because of the way preact handles refs, we can't be sure childData will be present for a particular key.  This fixes crashes, and things seem to work fine.

I'm a bit nervous I am just introducing subtle bugs when using preact instead of actually fixing it, but I don't really understand the code base.  Code review would be appreciated.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,6 +1,3 @@
-/* eslint-disable */
-const path = require('path');
-
 module.exports = {
   resolve: {
     alias: (process.env.REACT_IMPL === "preact") ? {
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,0 +1,23 @@
+/* eslint-disable */
+const path = require('path');
+
+module.exports = {
+  module: {
+    loaders: [
+      {
+        test: /css$/,
+        loaders: [ 'style', 'css' ],
+        include: path.resolve(__dirname, '../')
+      }
+    ]
+  },
+
+  resolve: {
+    alias: (process.env.REACT_IMPL === "preact") ? {
+      "react": "preact-compat",
+      "react-dom": "preact-compat",
+      "create-react-class": "preact-compat/lib/create-react-class",
+    } : {
+    }
+  }
+}
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -2,16 +2,6 @@
 const path = require('path');
 
 module.exports = {
-  module: {
-    loaders: [
-      {
-        test: /css$/,
-        loaders: [ 'style', 'css' ],
-        include: path.resolve(__dirname, '../')
-      }
-    ]
-  },
-
   resolve: {
     alias: (process.env.REACT_IMPL === "preact") ? {
       "react": "preact-compat",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "test": "./node_modules/.bin/karma start --single-run",
     "test:dev": "karma start karma.conf.js",
     "storybook": "start-storybook -p 9001",
-    "storybook:preact": "REACT_IMPL=preact start-storybook -p 9001"
+    "storybook:preact": "REACT_IMPL=preact npm run storybook"
   },
   "author": "Joshua Comeau <joshwcomeau@gmail.com>",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "babel-preset-stage-0": "6.1.18",
     "chai": "3.5.0",
     "create-react-class": "15.5.2",
-    "css-loader": "0.23.1",
     "eslint": "3.10.0",
     "eslint-config-airbnb": "12.0.0",
     "eslint-plugin-flowtype": "2.32.1",
@@ -87,7 +86,6 @@
     "react-dom": "15.5.4",
     "sinon": "1.17.3",
     "sinon-chai": "2.8.0",
-    "style-loader": "0.13.1",
     "webpack": "1.12.11"
   }
 }
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "prepublishOnly": "npm run build",
     "test": "./node_modules/.bin/karma start --single-run",
     "test:dev": "karma start karma.conf.js",
-    "storybook": "start-storybook -p 9001"
+    "storybook": "start-storybook -p 9001",
+    "storybook:preact": "REACT_IMPL=preact start-storybook -p 9001"
   },
   "author": "Joshua Comeau <joshwcomeau@gmail.com>",
   "license": "MIT",
@@ -79,6 +80,8 @@
     "karma-webpack": "1.7.0",
     "lodash": "4.3.0",
     "mocha": "2.4.5",
+    "preact": "^8.1.0",
+    "preact-compat": "^3.16.0",
     "prop-types": "15.5.8",
     "react": "15.5.4",
     "react-dom": "15.5.4",
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -66,6 +66,8 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
   // its job. We store these as a property on the instance. We're not using
   // state, because we don't want changes to trigger re-renders, we just
   // need a place to keep the data for reference, when changes happen.
+  // This field should not be accessed directly. Instead, use getChildData,
+  // putChildData, etc...
   childrenData: {
     /* Populated via callback refs on render. eg
     userSpecifiedKey1: {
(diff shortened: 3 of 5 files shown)


Hypnosphi:
>the way preact handles refs

Is it documented anywhere? Maybe your fallback can be done just once, in the ref handler?

src/FlipMove.js
```diff
@@ -479,7 +479,11 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
       }
 
       elements.push(child);
-      domNodes.push(this.childrenData[childKey].domNode);
+
+      const childData = this.childrenData[childKey]
+      if (childData) {
+        domNodes.push(childData.domNode);
+      }
```
Hypnosphi: please run `npm run build` before commiting =)

joshwcomeau:
Hey @AlexanderOtavka :wave:

Yeah, so I don't think this would introduce subtle bugs; it may actually prevent some edge-case bugs that I haven't had the time to dig into (specifically with rapid interrupts of enter/leave)

> Maybe in this case, there's a sense in extracting a method like that:

Hm, so I like the idea of this behaviour being abstracted, to help prevent bugs when updating the code. I don't think it should also set undefined values to an empty object, though, since it feels a bit wrong to me that a getter would have side effects.

As simple as it is, it could just be:

```js
// placed at the top of 
getChildDataByKey(key: string): ChildData {
  return this.childrenData[key] || {};
}
```

@AlexanderOtavka so right now our tests are miserable (my fault. I'd like to improve them soon), but you can do manual testing using the stories. Run `yarn storybook` and go through the list, checking visually for any errors as well as keeping an eye on the console. If all of the stories still work (except the ones that are specified as "invalid" or in Github Issues), I think we can be reasonably confident that nothing broke :)

src/FlipMove.js
joshwcomeau: For clarification, this is because there's a `prebuild` script that will run the linter, flow checker, and tests. It'll catch the missing semicolon that's causing CI to fail.

(Thanks again for adding that script, @Hypnosphi! And thanks for this review)

Hypnosphi:
>it feels a bit wrong to me that a getter would have side effects

You're right, but Maps in different languages (e.g. [Kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/get-or-put.html)) offer some `getOrPut`/`getOrSet`/`getOrAdd` method, which is quite useful. So this empty object could just be just passed explicitly. This would help in cases where we mutate the `childData` object (otherwise we'd mutate a detached empty object).

joshwcomeau:
> You're right, but Maps in different languages (e.g. Kotlin) offer some getOrPut/getOrSet/getOrAdd method, which is quite useful. So this empty object could be just passed explicitly. This would help in cases where we mutate the childData object (otherwise we'd mutate a detached empty object).

Hmmm, yeah. If we don't have a given child, though, I don't think we should be mutating it. If the `getOrPut` replaces it with an empty object, then `this.childrenData[key]` would return an empty object, which is truthy, when really I think we want to return `undefined` so that we can bail early.

This is very high-level to me, though, as I don't have the time atm to look through and see where we mutate `this.childrenData`. I'll give this a deeper look-through hopefully soon.

Hypnosphi:
Well, at the moment there are only two such places:
https://github.com/joshwcomeau/react-flip-move/blob/master/src/FlipMove.js#L642
https://github.com/joshwcomeau/react-flip-move/blob/master/src/FlipMove.js#L525-L529

In the former case, we're doing effectively `getOrPut(this.childrenData, key, {})` a couple lines before, and in the latter we check for `domNode` presence

Hypnosphi:
@AlexanderOtavka how to reproduce the crash, by the way?

src/FlipMove.js
AlexanderOtavka: Good to know!

AlexanderOtavka:
@Hypnosphi You should be able to repro just by using flip move with preact-compat, and then deleting an element.  It consistently errors out for me.

Hypnosphi:
I did it in [a branch](https://github.com/Hypnosphi/react-flip-move/tree/feature/storybook-preact) and there are no errors when removing

AlexanderOtavka:
@Hypnosphi I just checked it myself with story book and you are correct.  It looks like the issue is more complicated than just preact.  I'm using react-router as well as react-infinte-scroller and those might be affecting it in combination with preact.

src/FlipMove.js
```diff
@@ -599,6 +608,25 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
...
+  getChildData(key: string): NodeData {
+    return this.childrenData[key] || {};
+  }
+
+  putChildData(key: string, data: NodeData): void {
+    this.childrenData[key] = {
+      ...this.childrenData[key] || {},
+      ...data,
+    };
+  }
```
Hypnosphi: Good solution

package.json
```diff
@@ -79,6 +80,8 @@
     "karma-webpack": "1.7.0",
     "lodash": "4.3.0",
     "mocha": "2.4.5",
+    "preact": "^8.1.0",
```
Hypnosphi: Where is it used?

AlexanderOtavka: Used in `.storybook/webpack.config.js` for the `storybook:preact` script in `package.json`.

Hypnosphi: I see only `preact-compat` there

.storybook/webpack.config.js
```diff
@@ -0,0 +1,23 @@
...
+module.exports = {
+  module: {
+    loaders: [
+      {
+        test: /css$/,
+        loaders: [ 'style', 'css' ],
+        include: path.resolve(__dirname, '../')
+      }
+    ]
+  },
```
Hypnosphi: This loader can be deleted safely, we don't use any stylesheets

package.json
joshwcomeau: Yeah, it looks to me like `preact` isn't being used anywhere?

src/FlipMove.js
```diff
@@ -599,6 +608,25 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
...
+    return Object.prototype.hasOwnProperty.call(this.childrenData, key);
+  }
+
+  getChildData(key: string): NodeData {
+    return this.childrenData[key] || {};
+  }
+
+  putChildData(key: string, data: NodeData): void {
+    this.childrenData[key] = {
+      ...this.childrenData[key] || {},
```
joshwcomeau: I think this needs to be changed to `...(this.childrenData[key] || {})`.

The reason is that the spread operator is much lower precedence than the `||` operator ([source](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence))

Actually, I think this could be improved as:

```js
putChildData(key: string, data: NodeData): void {
  this.childrenData[key] = {
    ...this.getChildData(key),
    data,
  }
}
```

joshwcomeau: Also, this is a nit, but could we rename `putChildData` to `updateChildData`?

Could just be me, but a "put" to me feels like a hard replace, rather than a merge. "Update" indicates more clearly that "We're updating a field", not "we're replacing the old object with this new one"

Hypnosphi: maybe `set` as in `setState`?

joshwcomeau: `set` works for me!

.storybook/webpack.config.js
AlexanderOtavka: I'll leave that to another PR.  We can probably just leave it though, since we might want to add stylesheets later and it isn't hurting anything.

AlexanderOtavka:
@joshwcomeau clear to merge?

src/FlipMove.js
AlexanderOtavka: `set` it is!  Also, I'll fix this.

package.json
AlexanderOtavka: @joshwcomeau preact is a peer dependency of preact-compat

Hypnosphi: Oh.. that means that master is broken right now =(

.storybook/webpack.config.js
```diff
@@ -1,10 +1,13 @@
-const reactImpl = process.env.REACT_IMPL;
-
-const alias = reactImpl ? {
-  "react": reactImpl,
-  "react-dom": reactImpl,
-} : {};
+/* eslint-disable */
+const path = require('path');
```
Hypnosphi: why?

package.json
```diff
@@ -21,7 +21,7 @@
     "test": "./node_modules/.bin/karma start --single-run",
     "test:dev": "karma start karma.conf.js",
     "storybook": "start-storybook -p 9001",
-    "storybook-preact": "REACT_IMPL=preact-compat npm run storybook"
+    "storybook:preact": "REACT_IMPL=preact start-storybook -p 9001"
```
Hypnosphi: Can we keep `npm run storybook` here so that it would be enough to change the storybook CLI params in one place when needed?

.storybook/webpack.config.js
AlexanderOtavka: My bad, I forgot we no longer need it.

AlexanderOtavka:
@joshwcomeau Does this look good to you?

PR #153, opened 2017-05-13
Add preact config for storybook

Hypnosphi:
This allows to discover some weird bugs, for example:

- run `yarn storybook-preact`
- open [an example](http://0.0.0.0:9001/?selectedKind=Appear%20Animations%20-%20composite&selectedStory=Default%20%28disabled%29&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel) with composite children
- press "Remove 1 Item"
- switch to a different example in the same group

As a result, one of the items is drawn far below its place and has `top: 173px` style attached


.storybook/webpack.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  resolve: {
+    alias: {
+      "react-impl": "react"
```
joshwcomeau: So, I'm curious why this is necessary.

For the `storybook-preact` compilation, it seems like we should alias `react` to either `preact-compat` or `preact`, whichever we want to test against. That way the stories themselves don't have to change, and the regular webpack is a bit simpler. If we decide we want to test `preact-compat` instead of `preact`, we just change the value in `.storybook-preact/webpack.config.js`, instead of having to update all the stories.

.storybook-preact/head.html
```diff
@@ -0,0 +1,5 @@
+<style>
+  html, body, #root {
+    height: 100%;
+  }
+</style>
```
joshwcomeau: Are the files in `.storybook-preact` the same as in `.storybook`, aside from the Webpack aliases?

If so, maybe there's a way we can use a single `.storybook` directory, but pass in some kind of argument from the CLI specifying which React lib to use:

```js
// Just an example, not sure how we'd get this value:
const reactLib = process.argv[2];

const aliases = {
  preact: {
      "react": "preact-compat",
      "react-dom": "preact-compat",
      "react-impl": "preact"
  },
  // Also not sure if a React alias is required (see further comment), but assuming it is:
  react: {...},
};

module.exports = {
  resolve: {
    alias: aliases[reactLib],
  },
};
```

.storybook/webpack.config.js
Hypnosphi: `preact` is used in examples components, and `preact-compat` is the substitution for React in libraries, such as storybook and FlipMove itself

.storybook-preact/head.html
Hypnosphi: Unfortunately, storybook doesn't forward arguments to webpack

Hypnosphi:
What do you mean by supporting `preact` vs `preact-compat`? Usually you can't use `preact` for external libraries

.storybook-preact/head.html
Hypnosphi: This can be achieved with env variables though

Hypnosphi:
Ok, now it's kinda included in #151

AlexanderOtavka:
Yeah, I ended up adding it to test on storybook there.

PR #154, opened 2017-05-13
Deduplicate warnings

Hypnosphi:
While working on #153, I've noticed then when preact children are supplied, they generate false positive warnings about them being SFC. This is not a really big issue, but, to stop spoiling the console, I've added some deduplication, so that each type of warning fires only once

### DIFF
--- a/src/error-messages.js
+++ b/src/error-messages.js
@@ -1,42 +1,52 @@
 // @flow
 import type { Presets } from './typings';
 
-export const statelessFunctionalComponentSupplied = () => `
+function warn(msg: string) {
+  let hasWarned = false;
+  return () => {
+    if (!hasWarned) {
+      console.warn(msg);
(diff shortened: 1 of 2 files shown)


src/error-messages.js
```diff
@@ -1,42 +1,52 @@
...
-export const statelessFunctionalComponentSupplied = () => `
+function warn(msg: string) {
+  let hasWarned = false;
+  return () => {
+    if (!hasWarned) {
+      console.warn(msg);
+      hasWarned = true;
+    }
+  };
+}
```
joshwcomeau: Nit: could we rename this to `warnOnce`?

Hypnosphi: Ok

joshwcomeau:
Published in `v2.9.10` yesterday, forgot to let you know!

PR #155, opened 2017-05-14
Hotfix: satisfy preact peerDependency to make storybook run

Hypnosphi:
As @AlexanderOtavka has [pointed out](https://github.com/joshwcomeau/react-flip-move/pull/151#discussion_r116388826), `preact` is a peer dep of `preact-compact`. In fact, storybook-preact doesn't run without it:

```
ERROR in ./~/preact-compat/dist/preact-compat.js
Module not found: Error: Cannot resolve module 'preact' in /Users/jetbrains/react-flip-move/node_modules/preact-compat/dist
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "karma-webpack": "1.7.0",
     "lodash": "4.3.0",
     "mocha": "2.4.5",
+    "preact": "8.1.0",
     "preact-compat": "3.16.0",
     "prop-types": "15.5.8",
     "react": "15.5.4",


AlexanderOtavka:
Did you run yarn to rebuild the lock file?  I don't see any changes there.  Otherwise looks good to me, although it is fixed in my PR.

joshwcomeau:
Yeah, looks good to me if we update the yarnfile!

Hypnosphi:
I did, but nothing has added because it was already [there](https://github.com/joshwcomeau/react-flip-move/blob/master/yarn.lock#L4604-L4606), left from the first version of #153

PR #158, opened 2017-05-16
Extract public API typings into libdef

Hypnosphi:
Should partially fix #157. The `react-flip-move_v2.9.x.js` file is exactly the same as the one commited in https://github.com/flowtype/flow-typed/pull/844, and they have to stay in sync. Ideally they shouldn't be touched though, unless a minor version comes and brings some new API

PR #160, opened 2017-05-20
Add typescript definition

### LINKED ISSUE #159: Add .d.ts typings file for TypeScript users
I've done a little research, and I can't find a way to generate .d.ts files from flowtype files, although going the other direction seems doable.  Ideally, we could generate these typings from our flowtype information so TypeScript users can get the same benefits of static typings and we wouldn't have to maintain separate typing information for TypeScript that could fall out of date.

Hypnosphi:
Fixes #159

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ lib
 dist
 examples
 flow-coverage
+.idea
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "2.9.10",
   "description": "Effortless animation between DOM changes (eg. list reordering) using the FLIP technique.",
   "main": "lib/index.js",
+  "typings": "typings/react-flip-move.d.ts",
   "files": [
     "dist",
     "lib",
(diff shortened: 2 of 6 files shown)


joshwcomeau:
Oh man I am in no position to be reviewing this, haha. I've never used Typescript.

Unless someone else sees this in the next few hours and has notes, I'm cool to merge and publish this along with #158. Because it doesn't touch the main src code, I'm confident it won't break anything for regular users. If Typescript users have issues with the typings, we can always patch it in subsequent releases.

vsaarinen:
Unfortunately, it looks like this doesn't work since the typings file is not included in the NPM package. `package.json` refers to `typings/react-flip-move.d.ts` but the `typings/` folder doesn't exist:
```shell
$ ls -l node_modules/react-flip-move/
total 40
-rw-r--r--   1 ville  staff  1088 May 10 14:50 LICENSE.md
-rw-r--r--   1 ville  staff  9447 May 25 02:31 README.md
drwxr-xr-x   4 ville  staff   136 May  4 12:37 dist
drwxr-xr-x  20 ville  staff   680 May 17 09:20 lib
-rw-r--r--   1 ville  staff  3222 May 25 03:27 package.json
drwxr-xr-x  12 ville  staff   408 May 17 09:20 src
```

Hypnosphi:
Oh, that's a shame . Thanks for noting that!

PR #164, opened 2017-06-01
Fix broken link

karlhorky:


joshwcomeau:
Thanks! I did originally have my own write-up for FLIP, not sure where it went o_O linking to the original Aerotwist post makes sense to me though.

karlhorky:
No problem, thanks for the project! I just linked to the same thing that's linked to from the `master` Readme, so I thought that would be fine.

joshwcomeau:
Ohhh, totally didn't realize this was for the `gh-pages` branch! 

Nonetheless, thanks for the fix :)

PR #165, opened 2017-06-01
Fix TypeScript typings

vsaarinen:
Version 2.9.13 completely broke react-flip-move for me as a TypeScript user. I'm not fully versed in how the different module/bundling systems work, but it seems that the included bindings exported the type definitions in a way that wasn't aligned with how the component is actually exported. The included test TypeScript file compiled ok, but once you tried to use the component in a TypeScript application, everything blew up at runtime.

Using the [export =](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) method in the type definitions fixes these problems. This PR is based on the [@types/react-flip-move package's typings](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-flip-move) with a few additional fixes (the addition of the `style` and `verticalAlignment` props) and a combination of tests from both definitions.

The tests can be run (compiled) by running the following command inside `typings/`:

```shell
$ tsc --jsx react ./test.tsx
```

**Note:** I'm personally not very experienced with publishing type definitions, so there may be some errors here. However, the @types/react-flip-move definitions have worked well for me so far, and everything seemed to work when I replaced the published NPM package's `react-flip-move.d.ts` file with the one included in this PR.

### DIFF
--- a/typings/react-flip-move.d.ts
+++ b/typings/react-flip-move.d.ts
@@ -1,71 +1,235 @@
-import * as React from 'react';
+// Type definitions for react-flip-move 2.9
+// Project: https://github.com/joshwcomeau/react-flip-move
+// Definitions by: Joey Hain <https://github.com/jmhain>,
+//                 Ville Saarinen <https://github.com/vsaarinen>
+// TypeScript Version: 2.3
 
-export interface Styles {
-    [key: string]: string;
+import { Component, ReactElement } from 'react';
+
+export as namespace FlipMove;
+export = FlipMove;
+
+declare class FlipMove extends Component<FlipMove.FlipMoveProps, any> { }
+
+declare namespace FlipMove {
+    type AnimationPreset = 'elevator' | 'fade' | 'accordionVertical' | 'accordionHorizontal' | 'none';
+
+    interface Animation {
+        from: Partial<CSSStyleDeclaration>;
+        to: Partial<CSSStyleDeclaration>;
+    }
+
+    interface Styles {
+        [key: string]: string | number;
+    }
+
+    type AnimationProp = AnimationPreset | Animation | boolean;
+
+    interface FlipMoveProps {
+        className?: string;
+
+        /**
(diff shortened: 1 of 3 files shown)


typings/react-flip-move.d.ts
```diff
@@ -1,71 +1,233 @@
...
+         * - The fourth element will start transitioning after 60ms.
+         *
+         * Similarly to staggerDurationBy, This effect is great for "humanizing" transitions and making them feel less
+         * robotic.
+         *
+         * **Protip**: You can make elements animate one-at-a-time by using an identical duration and staggerDelayBy.
+         *
+         * @default 0
+         */
+        staggerDelayBy?: number;
```
Hypnosphi: those 4 props are can be strings as well

typings/react-flip-move.d.ts
```diff
@@ -1,71 +1,233 @@
...
+
+        /**
+         * Control the appear animation that runs when the component mounts. Works identically to enterAnimation below, but
+         * only fires on the initial children.
+         */
+        appearAnimation?: AnimationPreset | Animation | boolean;
```
Hypnosphi: let's keep this extracted to a type

typings/test.tsx
```diff
@@ -1,40 +1,114 @@
+// tslint:disable:max-classes-per-file
+
 import * as React from 'react';
-import FlipMove from '..';
+import * as FlipMove from '..';
 
 function childHook(el: React.ReactElement<any>, node: HTMLElement) {}
 function childrenHook(els: Array<React.ReactElement<any>>, nodes: Array<HTMLElement>) {}
 
 // tslint:disable-next-line:no-unused-expression
```
Hypnosphi: this can be removed, I believe

vsaarinen:
@Hypnosphi Was there still something that should be fixed?

Hypnosphi:
No, looks good to me

PR #166, opened 2017-06-18
Isolate testcases from each other

Hypnosphi:
I've checked that each testcase passes in focus (`it.only(...)`) mode. I've also added some eslint rules to ensure that no testcase is left focused or skipped by accidence.

### DIFF
--- a/test/helpers/index.js
+++ b/test/helpers/index.js
@@ -13,9 +13,8 @@ export function getTagPositions(renderedComponent) {
   );
 
   // returns { a: ClientRect, b: ClientRect, c: ClientRect }
-  return ['a', 'b', 'c'].reduce((acc, key) => {
-    // eslint-disable-next-line no-param-reassign
-    acc[key] = outputTags.find(el => el.id === key).getBoundingClientRect();
-    return acc;
-  }, {});
+  return ['a', 'b', 'c'].reduce((acc, key) => ({
+    ...acc,
+    [key]: outputTags.find(el => el.id === key).getBoundingClientRect(),
+  }), {});
 }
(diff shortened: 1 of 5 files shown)


Hypnosphi:
I also reimplemented `getTagPositions`. Before, it relied on the fact of being called strictly after reversing the articles' order. Now it finds children by id instead of using children order.

This discovered one more subtle issue: Because of using `Array#reverse` directly, original articles was mutated which led to switching `a` and `c` elements in half of testcase. The test have been passing up to this point just by fortune: the only testcase that actually checked the ids also happen to be the first one that does the reversing stuff.

So now I just clone the array before reversing.

test/helpers/index.js
```diff
@@ -11,10 +11,11 @@ export function getTagPositions(renderedComponent) {
...
-  return {
-    a: tagA.getBoundingClientRect(),
-    b: tagB.getBoundingClientRect(),
-    c: tagC.getBoundingClientRect(),
-  };
+
+  // returns { a: ClientRect, b: ClientRect, c: ClientRect }
+  return ['a', 'b', 'c'].reduce((acc, key) => {
+    // eslint-disable-next-line no-param-reassign
+    acc[key] = outputTags.find(el => el.id === key).getBoundingClientRect();
```
joshwcomeau: I'd do this the functional no-mutation way:

```js
return ['a', 'b', 'c'].reduce((acc, key) => ({
  ...acc,
  [key]: outputTags.find(el => el.id === key).getBoundingClientRect(),
}), {});
```

I forget if we've enabled the object rest/spread babel transform... if not, then the existing solution is fine (also cool with me if you want to add it)

Hypnosphi: I'm definitely in

PR #169, opened 2017-06-26
Use enzyme for testing React output

Hypnosphi:


test/index.spec.js
```diff
@@ -1,11 +1,14 @@
-/* eslint-disable no-undef, react/prop-types, react/no-multi-comp, no-unused-expressions */
+/* global chai, expect, sinon */
```
joshwcomeau: Neat, wasn't aware there was a workaround to `no-undef` for stuff like this. Guessing ESLint provides this `global` helper?

Hypnosphi: [Yes](http://eslint.org/docs/user-guide/configuring)

PR #174, opened 2017-07-08
Add tests for runtime props conversion and checks

Hypnosphi:
Add warning for text children

### DIFF
--- a/src/prop-converter.js
+++ b/src/prop-converter.js
@@ -21,6 +21,7 @@ import React, {
 
 import {
   statelessFunctionalComponentSupplied,
+  textNodeSupplied,
   invalidTypeForTimingProp,
   invalidEnterLeavePreset,
   deprecatedDisableAnimations,
@@ -49,11 +50,12 @@ declare var process: {
   },
 };
 
-let nodeEnv: string;
-try {
-  nodeEnv = process.env.NODE_ENV;
-} catch (e) {
-  nodeEnv = 'development';
+function isProduction(): boolean {
+  try {
+    return process.env.NODE_ENV === 'production';
+  } catch (e) {
+    return false;
+  }
 }
 
 function propConverter(
@@ -76,10 +78,10 @@ function propConverter(
     };
 
     // eslint-disable-next-line class-methods-use-this
-    checkForStatelessFunctionalComponents(children: mixed) {
+    checkChildren(children: mixed) {
       // Skip all console warnings in production.
       // Bail early, to avoid unnecessary work.
-      if (nodeEnv === 'production') {
+      if (isProduction()) {
         return;
       }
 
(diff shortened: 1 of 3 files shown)


src/prop-converter.js
```diff
@@ -49,11 +50,12 @@ declare var process: {
...
-try {
-  nodeEnv = process.env.NODE_ENV;
-} catch (e) {
-  nodeEnv = 'development';
+function isProduction(): boolean {
+  try {
+    return process.env.NODE_ENV === 'production';
+  } catch (e) {
+    return false;
+  }
```
joshwcomeau: Awesome, good helper.

test/index.spec.js
```diff
@@ -228,25 +229,193 @@ describe('FlipMove', () => {
...
   });
 
-  describe('duration prop runtime checking', () => {
-    let wrapper;
-    beforeEach(() => {
-      wrapper = shallow(<FlipMove />);
+  describe('prop runtime checking and conversion', () => {
+    let errorStub;
+    let warnStub;
+    const SFC = () => null;
```
joshwcomeau: Huh! I remember that SFCs couldn't return `null`, but apparently that was fixed a year and a half ago @_@, before React 15 @_@

test/index.spec.js
```diff
@@ -228,25 +229,193 @@ describe('FlipMove', () => {
...
+      errorStub.reset();
+      warnStub.reset();
     });
+    after(() => {
+      errorStub.restore();
+      warnStub.restore();
+    });
+
+    it('doesn\'t run checks in production environment', () => {
+      const env = process.env;
```
joshwcomeau: Maybe this should be moved into the `before` block, and reset in the `afterEach`, similar to the console stubs? That way each test can provide whatever `NODE_ENV` they want without worrying

test/index.spec.js
```diff
@@ -228,25 +229,193 @@ describe('FlipMove', () => {
...
     });
 
-    it('applies an array prop and throws', () => {
-      wrapper.setProps({ duration: ['hi'] });
-      expect(consoleStub).to.have.been.called;
+    describe('timing props', () => {
+      it('applies a bogus string', () => {
+        shallow(<FlipMove duration="hi" />);
+        expect(errorStub).to.have.been.calledWith(`
+>> Error, via react-flip-move <<
```
joshwcomeau: Maybe we should import the error string constructors and use them? That way we don't have to worry about updating the messages in both places if/when they change.

Could also be smart to add simple tests for the error message factories themselves... although I'm not too worried about it, since it's so simple, just string interpolation.

Hypnosphi: By not importing those but writing out the expected output, we kinda test the factories as well

Issue #120, opened 2016-12-26
On chrome mobile after 2 animations with many leaves then react-flip-move gets stuck in state where many nodes hang about that should have been removed.


ohjames:
Using a recent version of mobile chrome go to [kchomp.co](kchomp.co) and touch the `worldnews` link at the top followed by the `news` link as quickly as possible.

From here if you click the "news" link again you should see that a bunch of the nodes related to `worldnews` remain in the DOM. I verified with remote debugging that these nodes are no longer in `props.children`. From here clicking further links in the header leads to more and more HTML elements being left around. It seems that `react-flip-move` has gotten into a state where it can no longer remove needed elements.

I have been unable to replicate the issue on desktop chrome, it seems specific to mobile chrome. Using the latest react-flip-move (`2.7.3`).

ohjames:
I downgraded the `react-flip-move` version on [kchomp.co](kchomp.co) to `2.6.5` and no longer see the issue, I'm thinking it might have been introduced with the `2.7` series.

ohjames:
I've tried some other versions: The bug manifests in `2.6.6` but animates differently after the stuck case. Version `2.7.1` manifests the bug more violently. `2.7.3` isn't as bad as `2.7.1` but seems slightly worse than `2.6.6` when it comes to this.

So I assume it was broken in `2.6.6`, then got worse and was partially fixed between `2.7.1` and `2.7.3`. Eventually I'll stick `kchomp` on `2.6.6` but I'll leave it on `2.7.3` for now so you can see the issue.

joshwcomeau:
Hi @ohjames,

Right. Interrupts have always been a little tricky, especially when an item that was in the middle of exiting is re-entered. 

I can't promise I'll get to fixing it soon, I have a lot of other things vying for my attention. For now, I'd keep using `2.6.5`. I'm gonna try and roll a series of small bug fixes into a version at some point in the next month or two, will include this one.

ohjames:
Thanks for the info!

[...]

PR #179, opened 2017-08-24
Workaround for Zombie DOM Nodes

tobilen:
this adresses https://github.com/joshwcomeau/react-flip-move/issues/120#issuecomment-301613739

i've added a story that reproduces the issue based off of @ConneXN good work. my workaround sets the z-index for elements starting their leaving-transition to -1, so they dont catch click events meant for the (visible) element below them.

this is *not* a solve, seeing as we still have zombie nodes, but it satisfies my current requirments. maybe it helps someone else find a real solution.

### DIFF
--- a/src/dom-manipulation.js
+++ b/src/dom-manipulation.js
@@ -169,6 +169,7 @@ export const removeNodeFromDOMFlow = (
 
   const styles: Styles = {
     position: 'absolute',
+    zIndex: -1,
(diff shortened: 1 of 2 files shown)


tobilen:
actually this is rubbish. its relies on there being more added elements than removed ones, still leaves clickable areas to the side of the list and generally doesnt solve the issue of zombie elements.

PR #180, opened 2017-08-24
Fix Issue120

tobilen:
Adressing https://github.com/joshwcomeau/react-flip-move/issues/120

So after spending some time tracking down the issue demonstrated in https://github.com/joshwcomeau/react-flip-move/pull/180/files#diff-5bc397e1b32daa8794012a8ea12ce91aR17 (based off of @ConneXNL work  ❤️ ), i found that upon deletion of the childData entry, state is not updated. this means no rerender is triggered, and when a rerender is eventually triggered, we still have a child without corresponding childData (i.e. a zombie).

i've console.logged the lifecycle of such an element in this commit: https://github.com/joshwcomeau/react-flip-move/pull/180/commits/7f6f5972a9141e9c29b2fd5bdbe137bcb2bcf49a

the fix implemented here updates state in the `removeChildData` call. this seems to be working fine, but i'd appreciate some feedback.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -420,7 +420,6 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
       if (ev.target !== domNode) return;
 
       if (getKey(child) === '.$item151')
-        console.log('triggering tranitionEnd listener for 151');
       // Remove the 'transition' inline style we added. This is cleanup.
       domNode.style.transition = '';
 
@@ -430,7 +429,6 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
       domNode.removeEventListener(transitionEnd, transitionEndHandler);
 
       if (child.leaving) {
-        if (getKey(child) === '.$item151') console.log('removing child 151');
         this.removeChildData(getKey(child));
       }
     };
@@ -620,14 +618,13 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
   }
 
   removeChildData(key: string): void {
-    if (key === '.$item151') console.log("removing item151 from this.childrenData")
     delete this.childrenData[key];
-    /*this.setState(prevState => {
+    this.setState(prevState => {
       return {
         ...prevState,
         children: prevState.children.filter(child => child.key !== key),
       };
-    });*/
+    });
   }
 
   createHeightPlaceholder(): Element<*> {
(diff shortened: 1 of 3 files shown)


joshwcomeau:
Cool, thanks for your work on this!! Don't have time to dig into this now, but I'll try to go through this weekend and give it a proper review.

Curious why so much of the diff appears to be style changes, though? While I happen to agree with many of the tweaks (this project is a couple years old now and so my conventions have become more community-aligned in that time), I'd like for those changes to be done in a separate PR, so they can be discussed/debated outside of the underlying fix.

tobilen:
> Curious why so much of the diff appears to be style changes, though?
muscle memory (i have prettier on a hotkey) :P

i've removed the style changes

Hypnosphi:
FWIW, I would welcome adding Prettier to our setup (in a separate PR, of course)

src/FlipMove.js
```diff
@@ -624,6 +624,10 @@ class FlipMove extends Component<void, ConvertedProps, FlipMoveState> {
 
   removeChildData(key: string): void {
     delete this.childrenData[key];
+    this.setState(prevState => ({
+      ...prevState,
```
joshwcomeau: I don't believe this line is necessary; I know it isn't for the traditional object `setState` call, but I believe the functional version is the same. It merges in what you return with the existing state.

joshwcomeau:
> FWIW, I would welcome adding Prettier to our setup (in a separate PR, of course)

Ditto. Been using Prettier on other projects and it's delightful.

joshwcomeau:
@tobilen this is released, in v2.9.15 🎉 

I've also sent an invitation to add you as a collaborator on this project. It's a policy I've seen others do with success, where once a high quality PR is landed, the author is invited to collaborate on the project. Collaborators can review code, merge PRs, etc. Note that there are no "strings" or anything, accepting the invitation doesn't mean you'll be tasked with anything. Totally up to you how much or how little you'd like to do :)

Feel free to review and merge any PRs that come in that you're confident won't break anything, but please don't merge any of your own work unless it's been reviewed.

Thanks again for your work on this!

src/FlipMove.js
tobilen: > prevState is a reference to the previous state. It should not be directly mutated. Instead, changes should be represented by building a new object based on the input from prevState and props
https://facebook.github.io/react/docs/react-component.html#setstate

im actually pretty sure that whatever you return overwrites anything in the store. how would you remove entries from the store otherwise? at the very least, it seems to be encouraged by facebook (and immutability design patterns) to return an entire state object

PR #183, opened 2017-09-02
Migrate to flow v0.54.0

Hypnosphi:
Flow has recently introduced a lot of improvements in field of React: https://flow.org/en/docs/react/

PR #185, opened 2017-09-02
Fix weak types in libdef

Hypnosphi:
This is a follow-up to #183 

It turned out that the way I used react types in libdef file leads to all of them coercing to `any` (see https://github.com/flowtype/flow-typed/issues/1209#issuecomment-326756417)

This PR fixes that, plus restricts children types to react elements, and items filtered out by `Children.toArray` (those include `undefined`, `null`, and booleans)

I also made a PR to `flow-typed` with new libdef, so that it's available for external usage: https://github.com/flowtype/flow-typed/pull/1210

### DIFF
--- a/flow-typed/react-flip-move_v2.9.x.js
+++ b/flow-typed/react-flip-move_v2.9.x.js
@@ -1,17 +1,15 @@
-import type { ComponentType, Element, Node } from 'react';
-
-declare module 'react-flip-move' {
+declare module "react-flip-move" {
   declare export type Styles = {
-    [key: string]: string,
+    [key: string]: string
   };
 
   declare type ReactStyles = {
-    [key: string]: string | number,
+    [key: string]: string | number
   };
 
   declare export type Animation = {
     from: Styles,
-    to: Styles,
+    to: Styles
   };
 
   declare export type Presets = {
@@ -19,7 +17,7 @@ declare module 'react-flip-move' {
     fade: Animation,
     accordionVertical: Animation,
     accordionHorizontal: Animation,
-    none: null,
+    none: null
   };
 
   declare export type AnimationProp = $Keys<Presets> | boolean | Animation;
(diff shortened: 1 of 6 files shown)


flow-typed/react-flip-move_v2.9.x.js
```diff
@@ -1,25 +1,23 @@
-import type { ComponentType, Element, Node } from 'react';
-
-declare module 'react-flip-move' {
+declare module "react-flip-move" {
   declare export type Styles = {
-    [key: string]: string,
+    [key: string]: string
```
joshwcomeau: Guessing trailing commas aren't kosher in libdefs?

flow-typed/react-flip-move_v2.9.x.js
```diff
@@ -30,56 +28,80 @@ declare module 'react-flip-move' {
...
+    Hooks & {
+      children?: ChildrenArray<Child>
+    };
+
+  declare export type FlipMoveProps = FlipMoveDefaultProps &
+    CommonProps &
+    DelegatedProps & {
+      appearAnimation?: AnimationProp,
+      disableAnimations?: boolean // deprecated, use disableAllAnimations instead
+    };
```
joshwcomeau: How does flow-typed work in terms of Flow versions? 

I believe in recent releases of Flow, this could be rewritten as:

```js
declare export type FlipMoveProps = {
  ...FlipMoveDefaultProps,
  ...CommonProps,
  ...DelegatedProps,
  appearAnimation?: AnimationProp,
  disableAnimations?: boolean
}
```

But yeah, maybe `flow-typed` aims to be compatible with earlier versions of Flow?

src/FlipMove.js
```diff
@@ -43,6 +44,12 @@ function getKey(childData: ChildData): Key {
   return childData.key || '';
 }
 
+function getElementChildren(children: ChildrenArray<Child>): Array<Element<*>> {
+  // Fix incomplete typing of Children.toArray
```
joshwcomeau: Is this just a limitation of the Children type defined in React? Could you elaborate on the conditions required to remove the eslint-disable?

src/prop-converter.js
```diff
@@ -90,7 +90,8 @@ function propConverter(
       // If the child doesn't have a key, it means we aren't animating it.
       // It's allowed to be an SFC, since we ignore it.
       Children.forEach(children, (child: Child) => {
-        if (child == null) {
+        // null, undefined, and booleans will be filtered out by Children.toArray
+        if (child == null || typeof child === 'boolean') {
```
joshwcomeau: Curious why the boolean check was moved here?

flow-typed/react-flip-move_v2.9.x.js
Hypnosphi: You just specify version range. Thanks for the hint, I'll try that

flow-typed/react-flip-move_v2.9.x.js
Hypnosphi: Don't think so

src/FlipMove.js
Hypnosphi: `React.Children.toArray` removes `null`, `undefined`, and booleans from children set. The current [typing](https://github.com/facebook/flow/blob/master/lib/react.js#L231) for it is:
```
toArray<T>(children: ChildrenArray<T>): Array<$NonMaybeType<T>>;
```

This rules out `null` and undefined, but booleans are left

src/prop-converter.js
Hypnosphi: Because it's OK to do things like
```
<FlipMove>
  {someFlag && <SomeChild/>}
  {otherFlag || <OtherChild/>}
</FlipMove>
```

flow-typed/react-flip-move_v2.9.x.js
Hypnosphi: If I try to make this replacement, my `flow-typed` [tests](https://github.com/Hypnosphi/flow-typed/blob/master/definitions/npm/react-flip-move_v2.9.x/test_react-flip-move_v2.9.x.js) fail with following (when testing against Flow `v0.54`):
```
* react-flip-move_v2.9.x/flow_v0.53.0- (flow-v0.54.0): Unexpected Flow errors(2):
   Error: 0-test_react-flip-move_v2.9.x.js:19
                       v------------------------
    19:   getPosition={(node: HTMLElement) => ({
    20:     top: 0,
    21:     right: 0,
   ...:
    26:   })}
          -^ function. Could not decide which case to select
                                              v
    95:   declare export type FlipMoveProps = {
    96:     ...FlipMoveDefaultProps,
    97:     ...CommonProps,
   ...:
   101:   };
          ^ object type. See lib: react-flip-move_v2.9.x.js:95
     Case 1 may work:
      62:     getPosition: GetPosition,
                           ^^^^^^^^^^^ GetPosition. See lib: react-flip-move_v2.9.x.js:62
     But if it doesn't, case 2 looks promising too:
                                                v
      95:   declare export type FlipMoveProps = {
      96:     ...FlipMoveDefaultProps,
      97:     ...CommonProps,
     ...:
     101:   };
            ^ object type. See lib: react-flip-move_v2.9.x.js:95
     Please provide additional annotation(s) to determine whether case 1 works (or consider merging it with case 2):
                                                 v
      19:   getPosition={(node: HTMLElement) => ({
      20:     top: 0,
      21:     right: 0,
     ...:
      26:   })}
            ^ return

```

Plus it doesn't work in `v0.53`, so If I do that, I'd have to leave a separate libdef for one minor version only

joshwcomeau: Yeah, totally fair. Wasn't aware that there was a difference! Thanks for trying it :)

PR #190, opened 2017-09-28
Add support for React 16

joshwcomeau:
With React's official launch, I figured it was time to update the peer dependencies!

While trying to publish, I ran into some trouble; it wasn't able to find `babel-cli` when running the `lib` build step :/

Upgrading the dependency seems to fix it (publish still failed, but only because I hadn't yet incremented the version). I figured before I published, though, I should get another pair of eyes.

Mainly I'm curious if there's a more-ideal way to add support for React 16 in the peer dependencies.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-flip-move",
-  "version": "2.9.14",
```
joshwcomeau: Guess I never pushed the `2.9.15` branch, my bad. That was the release for zombie-nodes bugfix.

Hypnosphi:
Did you test it with components returning strings or fragments?

Hypnosphi:
> a more-ideal way to add support for React 16 in the peer dependencies

How about wildcard?

package.json
```diff
@@ -46,14 +46,14 @@
     "web-animations"
   ],
   "peerDependencies": {
-    "react": "0.13.x || 0.14.x || 15.x.x",
-    "react-dom": "0.13.x || 0.14.x || 15.x.x"
+    "react": "0.13.x || 0.14.x || 15.x.x || 16.x.x",
```
tobilen: would `"react": ">=0.13.x <=16.x.x",` work?

Hypnosphi:
> `if (typeof this.props.children === 'string' || typeof this.props.children === 'array')`

It's not about `this.props.children`, it's about output of each child's `render`

tobilen:
check is not needed anymore when we merge https://github.com/joshwcomeau/react-flip-move/pull/191

joshwcomeau:
This is no longer needed thanks to @tobilen's fantastic work :D

PR #191, opened 2017-09-30
Upgrade to react 16

tobilen:
This kind of snowballed.

the old version of storybook doesnt work with react16. it was throwing exceptions before, so i just upgraded it to the most recent version.
This meant webpack had to be upgraded, so i've done that as well. The stories are all looking good so far.

The big problem with react 16 is enzyme. I imported the adapter, but chai-enzyme doesnt play nice with enzyme 3 yet. 
There is an open PR which solves most of the issues. the others i circumvented by refactoring the assertions a bit. until its merged, you need to build chai-enzyme yourself:
1. `cd node_modules/chai-enzyme`
2. `yarn && yarn build`

The `setProps()` call doesnt seem to be working the same way in enzyme 3, which made the `mountAttached` + `attachedWrapper.setProps(...)` logic in the tests break. i worked around this by remounting for every test, but this of course makes the old/new position comparisons not work.

=================================
Update:
Refactored testing component so it passes its state to flip-move instead of its props. This means we can use `setState(...)` instead of `setProps(...)` to trigger a transition.
I had to make an empty `setState(...)` call after initial mounting because ...well, setState updates in enzyme 3 still appear to be a bit wonky.
I also removed chai-enzyme assertions for the time being.


joshwcomeau:
Oh wow! Thanks so much for this, whole project is modernized :D

If I understand correctly:

1) `chai-enzyme` is incompatible with Enzyme 3, and Enzyme 3 is needed for React 16
2) There's an open PR on `chai-enzyme` (https://github.com/producthunt/chai-enzyme/pull/196?) which addresses this, but in the meantime the tests can pass if you build it yourself, since you've specified a fork right now.
3) The tests fail now because we can't use `setProps` anymore. Looking at the code, it seems like you're just mounting with the value, which means that the tests are failing (I presume because there's no longer a transition between two states, it just mounts at the final state?)

For those first two points, I wonder if chai-enzyme is worth the trouble. It's certainly a convenience, but I feel like regular Chai is good enough (as I see you've done, by just using `to.equal` instead of `to.have.id`).

For the third point... I initially thought this was an issue of state carrying between tests, but I remembered that @Hypnosphi's great work meant that we are already mounting for every `describe`. I ran the tests locally and it is a bit tough to tell what's going on, but yeah I think the issue is that we _need_ some way to update the props after the component has mounted; everything other than `appear` animations works on transitions between props.

So, i think we need to figure out what the deal is with `setProps`. Do you know more about how it changed / why it doesn't work anymore?

tobilen:
To your first point, correct. 
To your second point, not exactly. Using the fork means we get past [this line](https://github.com/joshwcomeau/react-flip-move/pull/191/files#diff-faa503f61f8addeb603095cc2f40c82eR17), but, like you've observed yourself, there dont seem to be many chai enzyme assertions that actually work. Removing them entirely might be more feasible.

Third point seems to be the biggest issue, i agree. I've read through the [enzyme migration guide](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md) a couple of times now, but i dont see anything that explains why `setProps` throws. The only viable workaround i see atm is defining a wrapper component that passes its state as props, and then just calling `setState` on the wrapper.

`setState` doesnt work like it did before, as i learned during the upgrade of our codebase at work yesterday. It doesnt throw, but it also doesnt always trigger lifecycle methods on children, but maybe we can mitigate that by calling `wrapper.update()`. I'll try that tomorrow

tobilen:
We're looking good now. I'd still like to add a story (and test) for passing fragments and string to flip-move though.

tobilen:
test for proper fragment rendering has been added ( https://github.com/joshwcomeau/react-flip-move/pull/191/commits/9022b85e6545628a3bec182383f6c1d54bece8ac ), i'd say we're good to go now.

.flowconfig
```diff
@@ -3,6 +3,8 @@
 .*/node_modules/editions/.*
 .*/node_modules/flow-coverage-report/.*
 .*/node_modules/preact/.*
+.*/node_modules/@storybook/.*
+.*/node_modules/radium/.*
```
joshwcomeau: Guessing modern versions of Storybook use Radium?

package.json
```diff
@@ -46,29 +46,29 @@
...
+    "react": ">0.13.x <16.x.x",
+    "react-dom": ">0.13.x <16.x.x"
   },
   "devDependencies": {
-    "@kadira/storybook": "2.30.1",
-    "@types/react": "15.0.24",
-    "babel": "6.1.18",
-    "babel-cli": "6.2.0",
-    "babel-core": "6.2.1",
+    "@storybook/react": "^3.2.8",
```
joshwcomeau: Nit: can we omit the `^`?

I don't really think this even matters anymore, with Yarn/NPM5. But, my inner OCD wants it to be consistent, haha.

webpack.config.production.js
```diff
@@ -8,6 +8,7 @@ config.plugins = [
   new webpack.DefinePlugin({
     'process.env.NODE_ENV': JSON.stringify('production'),
   }),
+  new webpack.optimize.ModuleConcatenationPlugin(),
```
joshwcomeau: Oooh, neat! Keen to see the effect this has on bundle size.

.flowconfig
Hypnosphi: Not directly, but via `react-treebeard`. We use `glamorous` for our own components

package.json
Hypnosphi: You probably also want to upgrade it to 3.2.12, there has been some critical bugfixes

package.json
```diff
@@ -46,29 +46,29 @@
     "web-animations"
   ],
   "peerDependencies": {
-    "react": "0.13.x || 0.14.x || 15.x.x",
-    "react-dom": "0.13.x || 0.14.x || 15.x.x"
+    "react": ">0.13.x <16.x.x",
```
Hypnosphi: screen shot 2017-10-05 at 00 53 52
http://jubianchi.github.io/semver-check/

.flowconfig
joshwcomeau: Oh cool! Wasn't aware you were a contributor to Storybook - thanks for your work there :)

package.json
joshwcomeau: Good catch :+1:

joshwcomeau: Looks like `>=0.13.x <=16.x.x` works

joshwcomeau:
Hm, so I pulled this locally, tried running storybook, and I get the following errors:

screen shot 2017-10-05 at 9 08 55 am

I tried removing `node_modules` and reinstalling, but it still seems like there's some React version discrepancy.

Have you encountered this? Any tips for how to get past it?

tobilen:
@joshwcomeau try upgrading to the latest version of yarn i guess https://github.com/storybooks/storybook/issues/1868

package.json
tobilen: 😊

tobilen:
@Hypnosphi can you update your review?

joshwcomeau:
Woooo!!! Psyched to have this ready :D can't thank you enough for your work on this!!

joshwcomeau:
Released in `v2.9.17`! 🎊 

(Also, sorry for not debugging the yarn issue more thoroughly myself; the issue was indeed using 0.x yarn, upgrading fixed it!)

PR #192, opened 2017-10-10
do absolute check for node type

AlexDunmow:
getNativeNode checks if element is instanceOf HTMLEment, which doesn't work if the FlipMove component is rendered inside an iFrame using React Portal.

Checked getNativeNode to check absolutely if node is a text node.


Hypnosphi:
The main change looks OK to me, but it's hard to find it in all the style changes. You probably need to tell your text editor to respect our eslint setup

AlexDunmow:
Will resubmit.

PR #193, opened 2017-10-10
do absolute check for node type

AlexDunmow:
getNativeNode checks if element is instanceOf HTMLEment, which doesn't work if the FlipMove component is rendered inside an iFrame using React Portal.

Changed getNativeNode to check absolutely if node is a text node.

This is a resubmission, stopped jsprettier from polluting file and added type refinement to satisfy flow.

### DIFF
--- a/src/dom-manipulation.js
+++ b/src/dom-manipulation.js
@@ -250,8 +250,8 @@ export const getNativeNode = (element: ElementRef<*>): ?HTMLElement => {
     // Text nodes are not supported
     return null;
   }
-
-  return foundNode;
+  // eslint-disable-next-line flowtype/no-weak-types
+  return ((foundNode: any): ?HTMLElement);
 };
 
 export const createTransitionString = (
--- a/src/dom-manipulation.js
+++ b/src/dom-manipulation.js
@@ -246,7 +246,7 @@ export const getNativeNode = (element: ElementRef<*>): ?HTMLElement => {
   // composite components.
   const foundNode: ?(Element | Text) = findDOMNode(element);
 
-  if (!(foundNode instanceof HTMLElement)) {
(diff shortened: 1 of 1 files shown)


Hypnosphi:
Looks like there are still some flow errors though

AlexDunmow:
`|| foundNode instanceof Text` didn't work unfortunately, had to use type casting through any:

```
// eslint-disable-next-line flowtype/no-weak-types
return ((foundNode: any): ?HTMLElement);
```

Hypnosphi:
> didn't work unfortunately
What was the error?

AlexDunmow:
```
Error: src/dom-manipulation.js:257
257:   return foundNode;
              ^^^^^^^^^ Element. This type is incompatible with the expected return type of
233: export const getNativeNode = (element: ElementRef<*>): ?HTMLElement => {
                                                             ^^^^^^^^^^^ HTMLElement
```

Hypnosphi:
Ok, so Element is more generic than HTMLElement. We could probably do some duck-typing based on HTMLElement [declaration](https://github.com/facebook/flow/blob/master/lib/dom.js#L1183), but I don't think it would be better than type casting

joshwcomeau:
Cool, thanks!

joshwcomeau:
Hey @AlexDunmow, this is released in `2.9.16`. Thanks again for your work on this!

I've sent you an invitation to add you as a collaborator. I've been giving contributors push access when they submit a high-quality PR, and yours certainly qualifies!

Of course it's 100% up to you how much work you want to put in; totally fine if you accept the invitation and then never touch the project :) I just want to make it easy for folks to contribute, if they want to.

The only rule is, please don't merge your own code unless I or one of the other collaborators has given it a ✅. Feel free to review/merge others' PRs, respond to issues, etc.

PR #194, opened 2017-10-11
Fix on IE10

nicolasraynaud-millevolts:
The `props` property doesn't seem to be initialized at this point when creating a <FlipMove> component, which leads to an error.

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -61,7 +61,7 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
   // of truth.
   state = {
     children: getElementChildren(
-      this.props.children,
+      this.props? this.props.children: [],
     ).map((element: Element<*>) => ({
       ...element,
       element,


joshwcomeau:
Huh, interesting. Seems odd to be that a variable's instantiation would be browser-specific o_O.

That said, I don't really see any downside to this fix, so if it works... ¯\_(ツ)_/¯

I have a couple style nits, but I can just do them as a followup commit. I'll share the commit here, so that if you're interested in contributing again, you'll know a bit more about our conventions.

nicolasraynaud-millevolts:
Thanks! Right after my PR I've seen that you set lint and styling in the pre-push hook, I'll PR correctly next time ;)

joshwcomeau:
Yeah, actually I'm surprised the hook didn't take care of the lint issue!

The change was small, just added a comment and fixed the ternary styling.

Thanks again :)

PR #210, opened 2017-12-22
Fix typescript typings and documentation section of typeName prop

egorshulga:
In readme it is mentioned that typeName accepts `null`. Previous typings did not allow to use it.

Also, there was a typo in documentation concerning typeName. Previously it was rendered like `null` was a default value.


joshwcomeau:
Thanks @egorshulga!

PR #211, opened 2018-01-03
Increasing performance by avoiding styles recalculations

egorshulga:
In my project I have faced performance issues: it took about 700ms to process animations for a list with 30 elements. I noticed styles recalculations took the biggest part of it.

image

It was mentioned (e.g. [here](http://gent.ilcore.com/2011/03/how-not-to-trigger-layout-in-webkit.html)) that DOM calculated sizes functions should be used with caution. To achieve highest performance reads and writes should be performed in batches.

After applying some changes time to rerender the list reduced by 4 times.

image

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -217,10 +217,10 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
     const childrenInitialStyles = dynamicChildren.map(child =>
       this.computeInitialStyles(child),
     );
-    dynamicChildren.forEach((child, n) => {
+    dynamicChildren.forEach((child, index) => {
       this.remainingAnimations += 1;
       this.childrenToAnimate.push(getKey(child));
-      this.animateChild(child, n, childrenInitialStyles);
+      this.animateChild(child, index, childrenInitialStyles[index]);
     });
 
     if (typeof this.props.onStartAll === 'function') {
@@ -386,11 +386,7 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
     });
   }
 
-  animateChild(
-    child: ChildData,
-    index: number,
-    childrenInitialStyles: Styles[],
-  ) {
+  animateChild(child: ChildData, index: number, childInitialStyles: Styles) {
     const { domNode } = this.getChildData(getKey(child));
     if (!domNode) {
       return;
@@ -404,7 +400,7 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
     // In FLIP terminology, this is the 'Invert' stage.
     applyStylesToDOMNode({
       domNode,
-      styles: childrenInitialStyles[index],
+      styles: childInitialStyles,
     });
 
     // Start by invoking the onStart callback for this child.
@@ -570,7 +566,7 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
       // It is possible that a child does not have a `key` property;
       // Ignore these children, they don't need to be moved.
       if (!childKey) {
-        childrenBoundingBoxes.push();
+        childrenBoundingBoxes.push(null);
         return;
       }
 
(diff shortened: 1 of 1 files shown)


src/FlipMove.js
```diff
@@ -213,10 +213,14 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
...
 
+    // Splitting DOM reads and writes to be peformed in batches
+    const childrenInitialStyles = dynamicChildren.map(child =>
+      this.computeInitialStyles(child),
+    );
     dynamicChildren.forEach((child, n) => {
       this.remainingAnimations += 1;
       this.childrenToAnimate.push(getKey(child));
-      this.animateChild(child, n);
+      this.animateChild(child, n, childrenInitialStyles);
```
joshwcomeau: Nit: I think we can just pass the specific style for this child, rather than the whole object, with `childrenInitialStyles[n]`, since the index `n` will be the same across that first map and the `forEach`.

(Also, we should really change `n` to be `index` to be consistent, but that's on me to tackle later. Sorry for the confusion with variable names!)

src/FlipMove.js
```diff
@@ -553,19 +561,24 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
...
+    // Splitting DOM reads and writes to be peformed in batches
+    const childrenBoundingBoxes = [];
+
     this.state.children.forEach(child => {
       const childKey = getKey(child);
 
       // It is possible that a child does not have a `key` property;
       // Ignore these children, they don't need to be moved.
       if (!childKey) {
+        childrenBoundingBoxes.push();
```
joshwcomeau: Nit: can we change this to `childrenBoundingBoxes.push(null);`? This could just be a subjective thing but I find it a bit clearer if we explicitly push a falsy value, rather than implicitly pushing `undefined`.

src/FlipMove.js
```diff
@@ -574,15 +587,28 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
...
         }),
+      );
+    });
+
+    this.state.children.forEach((child, index) => {
+      const childKey = getKey(child);
+
+      if (!childKey || !childrenBoundingBoxes[index]) {
+        return;
+      }
```
joshwcomeau: Nit: This is a super-nit, but I wonder if it makes sense to give that second new condition a variable name, to signify what it represents. Something like:

```js
const wasAbleToPopulateBoundingBox = !!childrenBoundingBoxes[index];

if (!childKey || !wasAbleToPopulateBoundingBox) {
  return;
}
```

egorshulga:
All your remarks are completely reasonable.

src/FlipMove.js
```diff
@@ -574,15 +583,30 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
...
           parentDomNode,
           getPosition: this.props.getPosition,
         }),
+      );
+    });
+
+    this.state.children.forEach((child, index) => {
+      const childKey = getKey(child);
+
+      const childBoundingBox = childrenBoundingBoxes[index];
```
joshwcomeau: Thanks for the variable name!

Could you quickly explain why it isn't being used in the early-return conditional anymore? I assume it's because `null` is a valid type for `NodeData.boundingBox`, and so if that's what the value is currently resolved as, it makes sense to set it to that? This makes sense to me, just wanna make sure this was an intentional omission and I'm not missing anything :)

Gonna accept the PR since it's not-blocking but if you happen to see this, please let me know!

egorshulga: It's just a small optimization because `childBoundingBox` would be `null` if `childKey` is falsey value (see line 568) and `childrenBoundingBoxes` have the same order as `children`.

egorshulga:
I have just a small question: when it will be released?

joshwcomeau:
> I have just a small question: when it will be released?

Shipped in `2.10.1` just now!

I realized during prep that I never shipped your previous contribution, fixing typescript typings and updating the docs. Very sorry about that, it must've slipped through the cracks during the holidays. But it's out now with `2.10.1` and I'll be sure to ship stuff that gets merged in sooner from now on!

Thanks again for your great contribution. I'll be sending you an invite to add you as a collaborator, which gives you full push/merge authorization. It doesn't mean that there are any expectations for you to contribute more to this project, I just wanna make it as low-friction as possible if you are interested :) feel free to review PRs from other folks (including merging them if they're high-quality), tackle/close issues, etc. Just please don't merge anything to master that hasn't been approved by someone else.

Issue #209, opened 2017-12-13
Function onFinishAll is never called


Tsourdox:
This is how I'm using it.
```
<FlipMove className="ui horizontal list" duration={300} onStartAll={this.onStartAll} onFinishAll={this.onFinishAll}>
   {this.items}
<FlipMove>
```

[onStart, onStartAll, onFinish] works as expected.
Using version 2.9.1

joshwcomeau:
Hm, sorry about that! Thanks for the heads-up! 

Don't have a ton of time on my hands for this ATM, PRs welcome!

tobilen:
i'll take a look at it at the end of the week. might also use it to do some dom cleanup, which hopefully prevents some more edge cases with nodes not being removed

tobilen:
tested this in chrome/ff/ie11 with v2.9.10 as well as 2.9.1 . handler is being invoked correctly.

@Tsourdox if the problem persists, im going to need a reproducable code snippet (for example a [codesandbox](https://codesandbox.io/s/new))

Tsourdox:
I believe I tested it in Safari but I'll se if I can test it again and se if the issue was on my behalf.

tobilen:
safari (11) works as well:
![image](https://user-images.githubusercontent.com/8217788/34650051-7d958f2a-f3ba-11e7-97dc-4bc39f00550f.png)

Tsourdox:
Tested it and it doesn't fire in any browser. I'm using FlipMove together with react-dnd but even if I remove that code the problem persists. I'll try and update to the latest version and se if that changes anything. At the moment I don't need the event so I'm good for now. Thx for looking in to it!

tobilen:
Ok then, closing this issue.
It could be a problem with another library stopping propagation of events, but i'm just speculating here. If you manage to reproduce the problem, just reopen it.

PR #212, opened 2018-01-05
Warn if children are disabled

tobilen:
In Internet Explorer 11 (and below), the `transitonend` event will not (consistently) fire for elements that have the `disabled` attribute. This means the callback for node removal is never fired and the supposedly removed nodes stay in the html.

This PR adds a new warning to inform about this quirk. Before a node is removed, the list children are checked for `disabled`. If one or more are found and animations are not disabled, the warning emits.

ps.: this also takes care of the cleanup issue i mentioned [here](https://github.com/joshwcomeau/react-flip-move/issues/209#issuecomment-354853118)

### DIFF
--- a/src/FlipMove.js
+++ b/src/FlipMove.js
@@ -13,7 +13,7 @@ import ReactDOM from 'react-dom';
 // eslint-disable-next-line no-duplicate-imports
 import type { Element, ElementRef, Key, ChildrenArray } from 'react';
 
-import { parentNodePositionStatic } from './error-messages';
+import { parentNodePositionStatic, childIsDisabled } from './error-messages';
 import './polyfills';
 import propConverter from './prop-converter';
 import {
@@ -347,6 +347,14 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
 
       leavingChildren.forEach(leavingChild => {
         const childData = this.getChildData(getKey(leavingChild));
+
+        // Warn if child is disabled
+        if (
+          !this.isAnimationDisabled(this.props) &&
+          childData.domNode &&
+          childData.domNode.disabled
+        )
+          childIsDisabled();
 
         // We need to take the items out of the "flow" of the document, so that
         // its siblings can move to take its place.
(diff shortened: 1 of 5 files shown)


src/FlipMove.js
```diff
@@ -348,6 +348,14 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
       leavingChildren.forEach(leavingChild => {
         const childData = this.getChildData(getKey(leavingChild));
 
+        // Warn if child is disabled
+        if (
+          !this.isAnimationDisabled(this.props) &&
+          childData.domNode &&
+          childData.domNode.disabled
+        )
+          childIsDisabled();
```
joshwcomeau: Nit: Can we wrap the conditional body in squiggly-brackets? As much as I like the terseness of bracket-free bodies, in my experience it's too easy to introduce subtle bugs. Plus, it's unconventional, and I try to keep with convention unless I have a really good reason to deviate.

I'm surprised Prettier didn't catch and rewrite this... is the hook firing properly?

tobilen: yes, i don't like bracketless (for multi line ifs) either. husky also executes correctly. prettier is cool with it

tobilen:
as a bonus i've added [cross-env](https://github.com/kentcdodds/cross-env) so the npm scripts can be executed on windows as well 😄

src/FlipMove.js
Hypnosphi: That's because you include `prettier` **preset**, which disables quite a lot of eslint rules, including `curly`:

https://github.com/prettier/eslint-config-prettier/blob/984de70e8c6b57684b444283561019389ccebd11/index.js

joshwcomeau:
Woo! Thanks for this, and for adding `cross-env`! Sorry it took me a couple days to circle back.

I was gonna ship it, but I'm having trouble getting it started on my work computer. Gonna set a reminder to do it on my home machine, should get a release out within a few hours.

joshwcomeau:
Released in `v2.10.2`!

PR #213, opened 2018-01-10
Revamped build infrastructure, added module entry, flat bundling

Andarist:
I'm also pretty sure that this is a bad practice - https://github.com/joshwcomeau/react-flip-move/blob/master/src/polyfills.js . You should never automatically patch user's environment - you can either force people to patch it themselves or just use those polyfills **locally** without messing global prototypes. I can provide a PR for this if you agree with me - just let me know.

**react-flip-move.min.js** - gzipped size
pre PR - 6697 bytes
post PR - 5100 bytes

Savings - nearly 24% 💰


tobilen:
thanks for this! any particular reason why you used rollup instead of parcel?

Andarist:
Not yet have experimented with parcel and I know rollup pretty well. That being said I've viewed parcel as more of a "complete bundler" (which also from what I know takes .html as its entry), while rollup is only a JS bundler, it wont bundle your css or any other assets - its scope is just more limited. Would have to explore if parcel is even a match for this kind of task.

tobilen:
fair enough

package.json
```diff
@@ -2,21 +2,19 @@
...
     "typings"
   ],
   "scripts": {
     "lint": "eslint .",
     "lint:fix": "npm run lint -- --fix",
-    "build:lib": "babel src --out-dir lib",
-    "build:umd": "webpack src/index.js dist/react-flip-move.js --config webpack.config.production.js",
-    "build:umd:min": "webpack src/index.js dist/react-flip-move.min.js --config webpack.config.production.min.js",
-    "build": "npm run build:lib && npm run build:umd && npm run build:umd:min",
+    "build": "rollup -c",
```
joshwcomeau: Curious why you changed `build:lib`? Seems like you'd want to leave the files unbundled for the NPM-distributed one, and only change the UMD build.

If I import this library in a webpack-built application, webpack will be bundling this library regardless; is the idea that because Rollup concatenates it into a single file, it would still be smaller in the host application, because it only has 1 webpack-module-wrapper chunk of code?

I also kinda like the idea of keeping the `lib` files as the ones being used in dev, so that you can poke around with the source in `node_modules` and not have to deal with one big file... although I suppose that's a pretty niche concern.

rollup.config.js
```diff
@@ -0,0 +1,86 @@
...
+    plugins: umdConfig.plugins.concat(
+      replace({
+        'process.env.NODE_ENV': JSON.stringify('production'),
+      }),
+      uglify({
+        compress: {
+          pure_getters: true,
+          unsafe: true,
+          unsafe_comps: true,
+          warnings: false,
```
joshwcomeau: what are `unsafe_comps` and `warnings`? Couldn't find it in the Uglify docs

src/FlipMove.js
```diff
@@ -770,4 +770,6 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
   }
 }
 
-export default propConverter(FlipMove);
+const enhancedFlipMove = /* #__PURE__ */ propConverter(FlipMove);
+
+export default enhancedFlipMove;
```
joshwcomeau: Curious about the changes to this file. What's `/* #__PURE__ */`?

src/prop-converter.js
```diff
@@ -47,14 +47,6 @@ declare var process: {
   },
 };
 
-function isProduction(): boolean {
-  try {
-    return process.env.NODE_ENV === 'production';
-  } catch (e) {
-    return false;
-  }
-}
```
joshwcomeau: Nice, happy we get to kill this. Is the idea that rollup automatically replaces `process.env` values?

tobilen:
> I wish there was a way to conditionally import code

i havent actually tried it, but webpacks import function should theoretically serve that purpose?
```js
if (true) {
  import('react').then(() => console.log('react was imported'))
}
```
https://webpack.js.org/guides/code-splitting/#dynamic-imports

joshwcomeau:
> i havent actually tried it, but webpacks import function should theoretically serve that purpose?

Right, but I wouldn't want the dependency to load async (since that means that the host app needs to have support for dynamic imports, plus it just doesn't seem like something that should be done async, since the polyfills are tiny)

really what I want is:

```js
try {
  require('react-flip-move-polyfills');
} catch (e) {
  // do nothing, not a real error
}
```

I believe I remember that requires can't be try/catched though? Maybe I'm misremembering. Also, I believe `import` statements (non-dynamic) can't be nested in any blocks, so they're automatically a non-starter?

Hypnosphi:
Please don't forget to bump major version when you merge and release this

tobilen:
correct on both counts i believe. there are some workarounds for trying require, but nothing solid enough so i would feel comfortable shipping it in a public library.

we have yet to come up with a *solid* solution for conditional / dynamic imports after not adopting systemJS. i was merely pointing to current attempts :)

src/FlipMove.js
tobilen: i think its an uglify magic comment. means it can strip out the function if `enhancedFlipMove` is not used anywhere. https://davidwalsh.name/compress-uglify (search for pure_func )

package.json
tobilen: i do it all the time 👼

src/prop-converter.js
tobilen: i figured the function doesnt get stripped, whereas the conditional block would be?

package.json
Andarist: > If I import this library in a webpack-built application, webpack will be bundling this library regardless; is the idea that because Rollup concatenates it into a single file, it would still be smaller in the host application, because it only has 1 webpack-module-wrapper chunk of code?

That's the reason - compiling into a "flat bundle" (single file output) will result in overall smaller size in the host application.

rollup.config.js
Andarist: from the UglifyJS Readme.md

> unsafe_comps (default: false) -- Reverse < and <= to > and >= to allow improved compression. This might be unsafe when an at least one of two operands is an object with computed values due the use of methods like get, or valueOf. This could cause change in execution order after operands in the comparison are switching. Compression only works if both comparisons and unsafe_comps are both set to true.

warnings can be removed though as `false` is actually the default

> warnings (default: false) -- display warnings when dropping unreachable code or unused declarations etc.

src/FlipMove.js
Andarist: Exactly, it means that this can be safely dropped if its result is not used anywhere. It doesnt matter for the library itself (as we are sure that this result is used within the library), might slightly matter if this library gets used in application or other library.

By potentially dropping (but ofc only if `enhancedFlipMove` is not used) this call (`propConverter(FlipMove)`) it will allow the minifier to drop the whole lib (`FlipMove` etc) as it wont be referenced anymore.

src/prop-converter.js
Andarist: > Nice, happy we get to kill this. Is the idea that rollup automatically replaces process.env values?

Yeah, that's done by `rollup-plugin-replace` for umd builds, cjs and es stays untouched in this regard and will most likely be replaced by consumer's bundler (webpack or whatever). This is common practice to branch code depending on the environment, react itself does this so users of `react-flip-move` should have their replacing plugins already in place.

And as mentioned branchinc code cant be dropped because of this function call, whereas when this gets replaced to i.e. smth like:
```js
if ("production" === "development") {
  // some dev only warnings
}
```

UglifyJs will be able to actually recognize it as `if (false)` and will drop the entire block as its unreachable.

Andarist:
> I believe I remember that requires can't be try/catched though? Maybe I'm misremembering. Also, I believe import statements (non-dynamic) can't be nested in any blocks, so they're automatically a non-starter?

The ideal goal is to provide a static structure (so no dynamic `require`/`import` calls) - which is easier to udnerstand by the tools and later i.e. removed.

The whole thing about polyfills from me starts with the fact that I write code that gets embedded on the websites and believe me - many sites across the internet are really not that great. I just use "dont mess with globals" really seriously and a rule of thumb, because it has caused us issues several times already. We have also found out that "fixing" people's environment is not the best thing to do, because people do lots of crazy stuff on their sites and they actually might rely in some weirdo and convoluted (believe me, it happens) ways on their **bugs**.

Im not 100% sure, but also those polyfills here might not be 100% spec compliant (spec is specifying behaviour for many edge cases that are just hard to get right with a simple "working" code) and one might be polyfilling later (after your code gets executed) with more spec compliant version, but also conditionally - so he'd code would actually never run because you have done his job for him and that might lead to some bizarre bugs too.

By removing polyfills from the source:
- you'd be "safer" for users
- you'd be able to trim down the size even further, because actually you dont have to worry about some stuff implemented in those polyfills

let's take such `every`, it can be expressed with:
```js
export const every = (predicate, arr) => {
  for (let i = 0; i < arr.length; i++) {
    if (!predicate(arr[i], i, arr)) {
      return false;
    }
  }
  return true;
};
```

and the rest of this code doesnt have to live in this repo.

I've added a wip commit to this PR (can be reverted if you dont like it) to see what it would give us here and the result after minification and gzipping is now: **4958**. Maybe not the biggest gain (but its still 3% from the previous result) in this regard - true gain in my opinion is not patching the user's environment.

Hypnosphi:
If yo decide to remove polyfills, please don't forget to bump major version

src/FlipMove.js
joshwcomeau: ohhh, awesome! Thanks for the explanation.

src/prop-converter.js
joshwcomeau: Cool! Awesome, thanks. Yeah, was sad that we lost dead code elimination with the try/catch, but it was necessary with the old infra.

package.json
joshwcomeau: Ok! Cool. Truthfully my feelings are still a bit mixed, but ultimately I think the size benefit outweighs the small DX cost. This does appear to be conventional with other libs, noticed styled-components does this.

Thanks for being patient with my questions about this!

rollup.config.js
joshwcomeau: :+1: thanks. Not sure how I missed it, must've been looking at different docs. Both settings makes sense!

package.json
Andarist: Not a problem, Im happy to answer any questions regarding those topics. For better DX we could emit sourcemaps too, but actually Im not a big fan of them as they obscure identifiers names and checking values of various expressions quickly in the console becomes much harder. Your call on all of this anyway :)

Andarist:
Can work on those failing tests later too, from what Ive seen in the CI logs they seemed unrelated, but who knows - better to fix them anyway.

BTW. before merging I would like to review my own work once more to fix some linting/typing issues. I've just made this in a form of WIP so its not merge-ready yet (not lacking much though ;) )

joshwcomeau:
> Can work on those failing tests later too, from what Ive seen in the CI logs they seemed unrelated, but who knows - better to fix them anyway.

I think the only thing that needs to be done for the tests are a couple lint/prettier fixes for the new helpers that replace the polyfills :)

The previous commit had some legit test failures, but those seem to have been fixed (at least, the latest commit passes on my machine, while the previous commit failed as it did in CI). So I think all that's needed is some lint/prettier cleanup.

Once those are patched, everything else looks good to me! Let me know when you're ready and we'll ship it 🚀

Andarist:
Ive fixed linting problems, tests are failing (still? again?), we'll try to look into this later today, but Im not sure how much time I will have

Andarist:
Ok, I think ive fixed tests. Let's see what CI has to say about that, but in general I think this is merge-ready. 

While you need to release this as major, its a good moment to drop deprecated stuff to downsize the lib even further :P

joshwcomeau:
You're 100% right; because this is a major release, I should go through and remove deprecations. I can think of at least 1 thing (the `accordian` typo presets, lol), but there may be others. I'll aim to have it shipped within 24h or so.

I'm also about to send an invite as a collaborator. Your contribution has certainly earned it! This gives you full write access to this repo, so feel free to respond/close issues, merge other folks' PRs, etc (naturally, the only rule is not to merge stuff in that hasn't been reviewed). That said, of course, there's no expectation for you to do any more work on the project! Use it as much or as little as you'd like :)

Andarist:
Thanks! I probably will turn off notifications, because Im involved in too many projects right now and I just can't handle more notifications than necessary 😥 You can always mention me though and I should respond

joshwcomeau:
Published as `3.0.0`! 😁

Thanks again for all your work on this @Andarist

Andarist:
Cool. Glad you like the changes :)

tobilen:
thanks for your awesome work @Andarist , thanks for being right on top of all this @joshwcomeau @Hypnosphi ❤️❤️. you guys rock

Issue #218, opened 2018-01-20
Typescript definition seams to have a wrong export


thupi:
Hi,

Typescript errors got me importing the component like `import * as FlipMove from 'react-flip-move';` :-) 

However, when compiled this seams to cause react to crash because the render methods can't handle object's but expects a function or class instead. This issue is gone when i import the component like `import FlipMove from 'react-flip-move'` but that won't pass the typeschecking :-) 

I was able to fix this by using `export defualt FlipMove` instead of `export = FlipMove` in the definition :-)

joshwcomeau:
Ah, sorry about that!

Sounds like a quick fix, but I've never used Typescript, so I'd be unable to verify that the fix works... I suppose that if there are unintended consequences or something, they can always be patched separately; as it's just a type definition and not runtime code I'm not super concerned about it.

If you or anyone can create a quick PR I'll be sure to get it landed/shipped; I'm currently juggling too many things and don't wanna commit to doing this myself, only to forget about it.

joshwcomeau:
actually... lemme just do it now, since it really is a 2-second thing

thupi:
@joshwcomeau Hi, 

No problem at all. I was able to solve the issue in a matter of minutes :)

I would not consider myself as a Typescript expert but at least i can try to take a look at the fix :-)

joshwcomeau:
Believe this was fixed in `3.0.1` :)

spcfran:
Hi guys,

Not sure this is completely fixed yet - It compiles and ultimately works in the browser, but typescript shows an error:

![image](https://user-images.githubusercontent.com/4398367/37719655-62c0da78-2d1d-11e8-929a-1e36a3a61217.png)

Any other flavour of import other than `import * as FlipMove from` gives a different typescript error and it doesn't even work on the browser.

I'm using react 16.2.0, react-flip-move 3.0.1 and typescript 2.7

joshwcomeau:
Ah, sorry to hear that! Reopening the issue.

Afraid I've never used Typescript, so I'm not the best person to try and troubleshoot :/ As a possible workaround, is it possible to ignore the typescript typings for specific third-party modules, so that it doesn't show up as an error in your project? With Flow, for example, it just assumes unrecognized imports are `any`, which means you lose the benefit of type safety for those modules, but it doesn't affect your project otherwise.

pinyin:
Same error here. `export = FlipMove` works for me.
[...]

PR #220, opened 2018-01-24
Fix typescript typing

joshwcomeau:
Quick fix for #218

@thupi if you could take a look and make sure I understood, that'd be great. Not a typescript user, so it's pretty important someone correct me if this is wrong :)

### DIFF
--- a/typings/react-flip-move.d.ts
+++ b/typings/react-flip-move.d.ts
@@ -7,7 +7,7 @@
 import { Component, ReactElement } from 'react';
 
 export as namespace FlipMove;
-export = FlipMove;
+export default FlipMove;
 
 declare class FlipMove extends Component<FlipMove.FlipMoveProps, any> { }
 


joshwcomeau:
Hmm, buncha tests failing as a result. Need someone with Typescript experience to take a look :)

thupi:
@joshwcomeau Hi, 

I took a quick look at your tests for Typings and discovered that you are importing the Component in the test using `import * as FlipMove from 'react-flip-move;` :-) I think the tests may pass if you correct the test in the same way i mentioned in the #218 :-) 

I hasn't been able to test it myself :-) Would yo be able to give it a shot and se if it passes :-) ?

vsaarinen:
I'll try to open a separate PR soon that should fix this.

joshwcomeau:
Thanks for the heads-up @thupi! Makes total sense.

And thanks for tackling this, @vsaarinen!

PR #222, opened 2018-01-29
Fix TypeScript export type

vsaarinen:
This should fix react-flip-move for TypeScript users. Replaces #220.


joshwcomeau:
Thanks for this!! :)

joshwcomeau:
Published in `v3.0.1`

PR #227, opened 2018-05-07
Update README to make it friendlier for beginners

joshwcomeau:
I realized that this project's README contained a lot of fairly
hostile language to folks without a ton of experience with React
and its ecosystem. I removed some of the problematic words like
"just", and added more info for folks to get started.

I'll give this a merge in a day or two unless anyone has objections!

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ React Flip Move
 
 This module was built to tackle the common but arduous problem of animating a list of items when the list's order changes.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,7 +131,7 @@ Curious how this works, under the hood? [__Read the Medium post__](https://mediu
 
 ### Wrapping Element
 
-By default, FlipMove wraps the children you pass it in a `<div>`:
+By default, Flip Move wraps the children you pass it in a `<div>`:
```
joshwcomeau: I noticed we were being inconsistent with `FlipMove` vs `Flip Move`, so I updated them all to `Flip Move`

PR #230, opened 2018-06-23
Add verticalAlignment description to API reference

everdimension:
I noticed this useful prop in the source code but there's no mention in the documentation


joshwcomeau:
Ah, good catch! I remember when we added this prop, I'm surprised we didn't think to add docs for it.

Thanks for this, looks great.

PR #238, opened 2018-10-18
Support styled-components 4

morleyzhi:
Previously, FlipMove would throw an error when its direct child was a `styled-components` 4 element. This tightens the SFC check to make sure it's actually a function. Fixes #238 

Also upgrade React and Enzyme to 16.5 so they both support styled-components' use of forwardRefs.


morleyzhi:
Fixes #238

src/helpers.js
```diff
@@ -41,7 +41,10 @@ export const isElementAnSFC = (element: Element<*>): boolean => {
     return false;
   }
 
-  return !element.type.prototype.isReactComponent;
+  return (
+    typeof element.type === 'function' &&
```
joshwcomeau: ahh good catch!

PR #239, opened 2019-01-28
add custom createTransitionString function

j3l11234:
In my case I want to add background-color transition. 
I think customization the createTransitionString function is a good way, Because we may add other transition.

``` css
{
  transition: transform 500ms ease-in-out 0ms, opacity 500ms ease-in-out 0ms, background-color 500ms ease-in-out 2000ms;
}
```


joshwcomeau:
hi @j3l11234,

Sorry I never addressed this. This project has been on the back burner for me, in a totally different kitchen. 

I'm not looking to add any more features to this repo, I don't have time to maintain it as-is. Glad that you've found functionality that is useful though! I'll leave this PR open so others can discover it / use your fork if they also need this.

PR #241, opened 2019-08-13
Update react, get rid of deprecation warnings

tobilen:
In order to be able to use flip-move with the upcoming react 17, as well as to get rid of deprecation warnings it currently throws, this pr:
* updates non-breaking dependencies
* updates the lowest working react version to 16.3
* uses the lifecycle methods prefixed with `UNSAFE_` (see https://reactjs.org/blog/2018/03/29/react-v-16-3.html#component-lifecycle-changes )

It would probably be better to refactor FlipMove for `useEffect` / `useLayoutEffect` hooks, but i currently do not have the time for this.

addendum: it also fixes deployment, as setting up a chrome on travis now works a littlebit different 🤪


shilangyu:
Please accept this PR @joshwcomeau, the warning is driving me crazy, thank you <3

tobilen:
@joshwcomeau can you publish a new version please?

src/FlipMove.js
```diff
@@ -394,6 +369,32 @@ class FlipMove extends Component<ConvertedProps, FlipMoveState> {
     });
   }
 
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps: ConvertedProps) {
```
httpsOmkar: I believe that prefix UNSAFE_ is not good option for long time. Instead we should something for long life.

tobilen:
for everybody who is waiting, i've published my fork here: https://www.npmjs.com/package/@invisionag/react-flip-move as a prerelease

Carduelis:
@tobilen thanks! 
From my understanding, you are still planning to update the current package, is it right?

joshwcomeau:
Shoot, sorry y'all. Totally missed this whole thing!

Just published `3.0.4` with this fix. Agree that it's a temporary stopgap, but yeah it does the trick for now! if anyone would like to update to hooks that would be 💯, but I don't currently have the time for this.

Thanks @tobilen!!

PR #245, opened 2020-03-25
Update gotchas with functional components

JPeer264:
It does work with functional components if they are used correctly.

```jsx
const MyComponent = forwardRef((_, ref) => (
	<div ref={ref} />
));
```


tobilen:
can you update as well in https://github.com/joshwcomeau/react-flip-move/blob/master/src/FlipMove.js#L733 ?

i'd also like it if you could extend this "gotcha" a bit into a "how to use flip move with functional components" section. at least a small example in the readme, a story in the storybook would be even better, if you're up for it.

JPeer264:
@tobilen as a lot of people are using functional components (I guess), I put the usage with functional components right next to the quickstart, to make it more prominent.

JPeer264:
Sure, no problem. Just ping me if there is something else to change

Issue #233, opened 2018-08-18
Looking for maintainers


joshwcomeau:
Hi folks!

I started working on React Flip Move about 2.5 years ago, in January of 2016. Since then, this library has become widely used, getting dowloaded over 100,000 times a month on NPM (!!!). That means about 1% of React downloads also include React Flip Move.

Unfortunately, since then, I've also started work on lots of other side-projects, and I feel like I'm spreading myself a bit too thin. My primary focus right now (in addition to my full-time day job) is on [Guppy](https://github.com/joshwcomeau/guppy), to make front-end web dev more approachable for folks from untraditional backgrounds.

If anyone is interested in carrying the torch for this project, it'd be super appreciated. Ideally, this would be one of Flip Move's current collaborators, or someone with experience managing an open-source project, but let's chat even if that doesn't apply to you.

For folks reading this and worrying because their app depends on React Flip Move, don't worry too much; even if we don't find a new maintainer, I'm going to try and provide baseline levels of support for this project. That said, at some point in the future, React might deprecate `findDOMNode`, and if they do, significant refactoring would need to happen to this project... so it may be wise to consider switching to something like [react-flip-toolkit](https://www.npmjs.com/package/react-flip-toolkit), which is actively under development.

xcv58:
Hi @joshwcomeau 

Thanks for this awesome library. I want to help, and I use React Flip Move in my chrome extension: https://github.com/xcv58/Tab-Manager-v2.

I use React, flow/TS, storybook in my daily job, and familiar with CI setup.

joshwcomeau:
Hi @xcv58! Thanks for reaching out :) super glad to have you onboard! Really appreciate it.

I'll send an invite to manage the GitHub repo. Feel free to respond to issues as they come up, or take a stab at fixing things! I'm happy to answer questions (although it's been a while since I've worked with this codebase myself...).

If things go well, I'll give you access to NPM publishing as well :)

mahaveer0496:
Hey @joshwcomeau  I would like to be a part if possible :)

ilyaskarim:
I'm also here to help this library. Let me know if need any help in a task.

joshwcomeau:
@xcv58 Ah, really sorry for being MIA! I don't check Github notifications because I get tagged in all sorts of stuff / it's 99% noise.

[...]

PR #268, opened 2022-04-21
Add explicit children prop

nickjs:
Hi!

`@types/react` has been updated for react 18 to no longer include an implicit `children` prop on every `Component`. This means `<FlipMove>` is no longer usable with the latest react typings because you must pass `children` to `<FlipMove>` but TypeScript thinks the `<FlipMove>` component doesn't accept any `children`.

This PR adds an explicit `children` prop to the typings including a doc comment.

Thanks!

### DIFF
--- a/typings/react-flip-move.d.ts
+++ b/typings/react-flip-move.d.ts
@@ -26,6 +26,11 @@ declare namespace FlipMove {
     type AnimationProp = AnimationPreset | Animation | boolean;
 
     interface FlipMoveProps {
+        /**
+         * The children to animate. Each child must have a unique key to work correctly.
+         */
+        children?: React.ReactNode;
+
         className?: string;
 
         /**


xcv58 (approved):
lgtm!

nickjs:
Thanks @xcv58!

xcv58:
> @xcv58 Can you create a release for this change? I am running into this issue with the latest version since it doesn't include this change

Sorry I don't have that permission to do the release, waiting for this https://github.com/joshwcomeau/react-flip-move/issues/233#issuecomment-1107257677

JakeThurman:
> > @xcv58 Can you create a release for this change? I am running into this issue with the latest version since it doesn't include this change
> 
> Sorry I don't have that permission to do the release, waiting for this https://github.com/joshwcomeau/react-flip-move/issues/233#issuecomment-1107257677

Saw that after posting! Thanks for taking the time to reply.

